PR #6, opened 2023-10-19
Create Rick & Morty API

Dimagaa:


src/main/java/mate/academy/rickandmorty/config/DataFetchConfig.java
```diff
@@ -0,0 +1,61 @@
...
+            HttpResponse<byte[]> response = httpClient.send(
+                    request,
+                    HttpResponse.BodyHandlers.ofByteArray()
+            );
+            return objectMapper.readValue(
+                    response.body(), new TypeReference<>() {
+                    }
+            );
+        } catch (IOException | InterruptedException e) {
+            throw new RuntimeException(e);
```
andrii-hoienko: It is better to throw custom exception instead of generic one

src/main/java/mate/academy/rickandmorty/config/DataFetchConfig.java
```diff
@@ -0,0 +1,61 @@
...
+public class DataFetchConfig {
+    private final CharacterService service;
+    private final ObjectMapper objectMapper;
+    @Value("${rickandmorty.api.url}")
+    private String characterUrl;
+
+    @EventListener(ApplicationReadyEvent.class)
+    public void fetchData() {
+        List<ExternalCharacterDto> characters = new ArrayList<>();
+        CharacterResponseDataDto response = loadData(characterUrl);
```
andrii-hoienko:
```suggestion
        CharacterResponseDataDto characterResponse = loadData(characterUrl);
```

src/main/java/mate/academy/rickandmorty/config/DataFetchConfig.java
```diff
@@ -0,0 +1,61 @@
...
+        do {
+            characters.addAll(response.results());
+            response = loadData(response.info().next());
+        } while (response.info().next() != null);
+        service.saveAll(characters);
+    }
+
+    private CharacterResponseDataDto loadData(String url) {
```
andrii-hoienko: OPTIONAL: As an alternative way to make http requests it is possible to use feignClient, which from my point of view is a more suitable and agile way of making web clients and request. But actually for current task your solution in ok as well

andrii-hoienko (changes requested):
GJ! Let`s improve your solution

vovasalyha (approved):

PR #8, opened 2023-10-19
Created API which fetches data from public API and has two methods

Someboty:
Created API which fetches data from public API and has two methods:
first gets a random character in the universe the animated series Rick & Morty; 
second takes a string as an argument, and returns a list of all characters whose name contains the search string.

### DIFF
--- a/src/main/java/mate/academy/rickandmorty/service/impl/CharacterClient.java
+++ b/src/main/java/mate/academy/rickandmorty/service/impl/CharacterClient.java
@@ -1,62 +1,59 @@
 package mate.academy.rickandmorty.service.impl;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import java.io.IOException;
-import java.net.URI;
 import java.net.http.HttpClient;
-import java.net.http.HttpRequest;
-import java.net.http.HttpResponse;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import lombok.RequiredArgsConstructor;
 import mate.academy.rickandmorty.dto.external.RequestDto;
 import mate.academy.rickandmorty.dto.external.RequestResultsDto;
-import mate.academy.rickandmorty.thread.CustomThread;
+import mate.academy.rickandmorty.thread.GetRequestDtoThread;
 import org.springframework.stereotype.Component;
 
 @Component
 @RequiredArgsConstructor
 public class CharacterClient {
-    private static final String BASE_URL = "https://rickandmortyapi.com/api/character/?page=";
     private static final int INITIAL_PAGE = 1;
 
     private final ObjectMapper objectMapper;
 
     public List<RequestResultsDto> getCharacters() {
         HttpClient httpClient = HttpClient.newHttpClient();
-        List<RequestResultsDto> characters;
-        int pages;
+        List<RequestResultsDto> characters = new ArrayList<>();
+        int pages = getPagesCountAndFirstPageResults(httpClient, characters);
+        completeCharacterList(httpClient, characters, pages);
+        return characters;
+    }
 
-        HttpRequest request = HttpRequest.newBuilder()
-                .GET()
-                .uri(URI.create(BASE_URL + INITIAL_PAGE))
-                .build();
-        try {
-            HttpResponse<String> response = httpClient.send(
-                    request, HttpResponse.BodyHandlers.ofString());
-            RequestDto dataDto = objectMapper.readValue(
-                    response.body(), RequestDto.class);
-            pages = dataDto.info().pages();
-            characters = new ArrayList<>(dataDto.results());
-        } catch (IOException | InterruptedException e) {
-            throw new RuntimeException(e);
-        }
+    private int getPagesCountAndFirstPageResults(
+            HttpClient httpClient, List<RequestResultsDto> characters) {
+        RequestDto firstPage =
+                new GetRequestDtoThread(httpClient, objectMapper, INITIAL_PAGE).call();
+        characters.addAll(firstPage.results());
+        return firstPage.info().pages();
+    }
+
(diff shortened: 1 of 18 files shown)


src/main/java/mate/academy/rickandmorty/service/impl/CharacterClient.java
```diff
@@ -0,0 +1,60 @@
...
+            RequestDto dataDto = objectMapper.readValue(
+                    response.body(), RequestDto.class);
+            pages = dataDto.info().pages();
+            characters = new ArrayList<>(dataDto.results());
+        } catch (IOException | InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+
+        for (int i = 2; i <= pages; i++) {
+            request = HttpRequest.newBuilder()
```
vovasalyha: [Optional] If you want to practice concurrency, you can spin up few threads to send requests in parallel (or even better, use CompletableFuture, or even Spring's WebClient which is by default reactive(asynchronous)). [Here's the link](https://www.baeldung.com/spring-5-webclient)

src/main/java/mate/academy/rickandmorty/service/impl/CharacterClientService.java
```diff
@@ -0,0 +1,22 @@
...
+public class CharacterClientService {
+    private final CharacterClient characterClient;
+    private final CharacterRepository characterRepository;
+    private final CharacterMapper characterMapper;
+
+    @PostConstruct
+    public void saveCharactersToDb() {
+        characterClient.getCharacters().stream()
+                .map(characterMapper::toEntity)
+                .forEach(characterRepository::save);
```
vovasalyha: Find out if this approach uses batch inserts by default, if no, then consider using saveAll() method

vovasalyha (approved):
GJ. Left few thoughts for improvement

src/main/java/mate/academy/rickandmorty/service/impl/CharacterClient.java
```diff
@@ -0,0 +1,72 @@
...
+            pages = dataDto.info().pages();
+            characters = new ArrayList<>(dataDto.results());
+        } catch (IOException | InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+
+        ExecutorService executorService = Executors.newFixedThreadPool(pages);
+        List<CompletableFuture<List<RequestResultsDto>>> futureRequests = new ArrayList<>();
+        for (int i = 2; i <= pages; i++) {
+            Future<List<RequestResultsDto>> future = executorService.submit(
```
vovasalyha: Your solution is mostly correct, but there is one important thing to keep in mind. You whether use ExecutorService or CompletableFuture. Using them both makes no sense, because using of executor service already runs your tasks in parallel. So you need to choose one of two options here:
- Use ExecutorService and collect List of Future then process results in a stream
```java
Future<List<RequestResultsDto>> future = executorService.submit(
            new CustomThread(httpClient, objectMapper, i));
    futureRequests.add(future);
```
```java
List<RequestResultsDto> additionalCharacters = futureRequests.stream()
        .map(future -> {
            try {
                return future.get();
            } catch (Exception e) {
                throw new RuntimeException(e);
            }
        })
        .flatMap(List::stream)
        .toList();
```
- Use CompletableFuture to submit tasks asynchronously (CompletableFuture uses embedded ForkJoinPool by default, which is created for each Java app)

```java
CompletableFuture.supplyAsync(() -> new CustomThread(httpClient, objectMapper, i).call())
...
```

But the general idea is correct, GJ.

vovasalyha (changes requested):

src/main/java/mate/academy/rickandmorty/service/impl/CharacterClient.java
```diff
@@ -0,0 +1,62 @@
...
+                    request, HttpResponse.BodyHandlers.ofString());
+            RequestDto dataDto = objectMapper.readValue(
+                    response.body(), RequestDto.class);
+            pages = dataDto.info().pages();
+            characters = new ArrayList<>(dataDto.results());
+        } catch (IOException | InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+        List<CompletableFuture<List<RequestResultsDto>>> futureRequests = new ArrayList<>();
+        for (int i = INITIAL_PAGE + 1; i <= pages; i++) {
```
andrii-hoienko: is it possible to use in IntStream.range (...

andrii-hoienko: I would prefer to add logic of receiving additionalCharacters to separate method

src/main/java/mate/academy/rickandmorty/thread/CustomThread.java
```diff
@@ -0,0 +1,37 @@
...
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.List;
+import java.util.concurrent.Callable;
+import lombok.RequiredArgsConstructor;
+import mate.academy.rickandmorty.dto.external.RequestDto;
+import mate.academy.rickandmorty.dto.external.RequestResultsDto;
+
+@RequiredArgsConstructor
+public class CustomThread implements Callable<List<RequestResultsDto>> {
```
andrii-hoienko: Not representative name

src/main/java/mate/academy/rickandmorty/thread/CustomThread.java
```diff
@@ -0,0 +1,37 @@
...
+    private static final String BASE_URL = "https://rickandmortyapi.com/api/character/?page=";
+    private final HttpClient httpClient;
+    private final ObjectMapper objectMapper;
+    private final int pageNumber;
+
+    @Override
+    public List<RequestResultsDto> call() {
```
andrii-hoienko: code duplication

andrii-hoienko (changes requested):
GJ! Let`s improve your solution

andrii-hoienko (approved):
GJ!

PR #12, opened 2023-10-20
Add working solution: created API with two methods: getRandomCharacte…

WasjaZeikan:
…r and getCharactersContainName using public Rick and Morty API

### DIFF
--- a/src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
+++ b/src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
@@ -64,10 +64,7 @@ public class CharacterServiceImpl implements CharacterService {
     }
 
     private void initCounts(HttpClient httpClient) throws IOException, InterruptedException {
-        HttpRequest httpRequest = HttpRequest.newBuilder()
-                .GET()
-                .uri(URI.create(String.format(URL,1)))
-                .build();
+        HttpRequest httpRequest = getCharacterRequest(1);
         HttpResponse<String> httpResponse = httpClient.send(httpRequest,
                 HttpResponse.BodyHandlers.ofString());
         CharacterMetaInformationResponseDto responseDto = objectMapper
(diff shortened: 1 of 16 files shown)


src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,89 @@
...
+        charactersCount = responseDto.getInfo().getCount();
+        pagesCount = responseDto.getInfo().getPages();
+    }
+
+    private CharacterEndpointResponseDto getCharactersFromApi(HttpClient httpClient, int page)
+            throws IOException, InterruptedException {
+        HttpRequest httpRequest = HttpRequest.newBuilder()
+                .GET()
+                .uri(URI.create(String.format(URL, page)))
+                .build();
```
koval-kn: it's better to exclude logic like that to separate method to avoid code duplication

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,89 @@
...
+                .stream()
+                .map(characterMapper::toDto)
+                .toList();
+    }
+
+    private void initCounts(HttpClient httpClient) throws IOException, InterruptedException {
+        HttpRequest httpRequest = HttpRequest.newBuilder()
+                .GET()
+                .uri(URI.create(String.format(URL,1)))
+                .build();
```
koval-kn: it's better to exclude logic like that to separate method to avoid code duplication

koval-kn (changes requested):

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,90 @@
...
+        HttpClient httpClient = HttpClient.newHttpClient();
+        try {
+            initCounts(httpClient);
+            for (int i = 1; i <= pagesCount; i++) {
+                characterRepository.saveAll(getCharactersFromApi(httpClient, i).getResults()
+                        .stream()
+                        .map(characterMapper::toModel)
+                        .toList());
+            }
+        } catch (IOException | InterruptedException e) {
```
Rommelua, line 45:
```suggestion
        } catch (Exception e) {
```

Rommelua (approved):

PR #18, opened 2023-10-23
Adds client for get and persist characters from API and controller for get random and by name characters.

BusyhinD:


Ivan95kos (approved):
Great job

PR #27, opened 2023-10-27
Feature/added api

Nikostrik:


src/main/java/mate/academy/rickandmorty/model/Character.java
```diff
@@ -0,0 +1,21 @@
...
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+import lombok.Data;
+
+@Data
+@Entity
+@Table(name = "characters")
```
koval-kn: to the future - conventionally we should name DB tables in singular, not plural:
```suggestion
@Table(name = "character")
```

src/main/java/mate/academy/rickandmorty/service/impl/CharacterClientImpl.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@RequiredArgsConstructor
+@Component
+public class CharacterClientImpl implements CharacterClient {
+    private final ObjectMapper objectMapper;
+    private final List<CharacterResponseDto> characterResponseDtoList = new ArrayList<>();
+    @Value("${mate.academy.rickandmorty.url}")
+    private String apiUrl;
```
koval-kn: if we'll change OpenAPI url, we'll get different response model, so we should change model in our  code too and we'll need to recompile it. so only reason to extract this url to the property is redundant. 
we can avoid extracting url to the property.

src/main/java/mate/academy/rickandmorty/service/impl/CharacterClientImpl.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    public List<CharacterResponseDto> getCharactersIntoDB() {
+        HttpClient httpClient = HttpClient.newHttpClient();
+
+        HttpRequest httpRequest = HttpRequest.newBuilder()
+                .GET()
+                .uri(URI.create(apiUrl))
+                .build();
```
koval-kn: extract to the separate method

src/main/java/mate/academy/rickandmorty/service/impl/CharacterClientImpl.java
```diff
@@ -0,0 +1,56 @@
...
+            HttpResponse<String> response = httpClient.send(httpRequest,
+                    HttpResponse.BodyHandlers.ofString());
+            CharacterInfoDto characterInfoDto = objectMapper
+                    .readValue(response.body(), CharacterInfoDto.class);
+            int count = Integer.parseInt(characterInfoDto.info().get("count"));
+            for (int i = 1; i <= count; i++) {
+                httpRequest = HttpRequest.newBuilder()
+                        .GET()
+                        .uri(URI.create(apiUrl + "/" + i))
+                        .build();
```
koval-kn: extract to the separate method

src/main/resources/application.properties
```diff
@@ -1 +1,10 @@
+spring.datasource.url=jdbc:mysql://localhost:3306/character_db?serverTimezone=UTC
+spring.datasource.username=root
+spring.datasource.password=124533
```
koval-kn: for future, better not expose credentials to the open-source git. google about how to avoid it (simplest way as for me - to leave properties with empty values to hide it from committing, commit it, push, and return them back)

koval-kn (changes requested):

koval-kn (approved):

PR #34, opened 2023-11-02
Add hw jv-rick-and-morty

vitaliiplys:


src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+@RequiredArgsConstructor
+public class CharacterServiceImpl implements CharacterService {
+    private static int countCharacter;
+    private static final Random RANDOM = new Random();
+    private final CharacterRepository characterRepository;
+    private final CharacterMapper mapper;
+
+    @Override
+    public CharacterDto getRandomCharacter() {
+        int id = RANDOM.nextInt(countCharacter) + 1;
```
olekskov: use nextLong() method and avoid casting int to long

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+import mate.academy.rickandmorty.dto.internal.CharacterDto;
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import mate.academy.rickandmorty.model.Character;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class CharacterServiceImpl implements CharacterService {
+    private static int countCharacter;
```
olekskov: redundant class variable, should be removed

src/main/java/mate/academy/rickandmorty/service/CharacterClientImpl.java
```diff
@@ -0,0 +1,37 @@
...
+    public List<CharacterResponseDto> getCharacters() {
+        HttpClient httpClient = HttpClient.newHttpClient();
+        HttpRequest request = HttpRequest.newBuilder()
+                .GET()
+                .uri(URI.create(CHARACTER_URL))
+                .build();
+        try {
+            HttpResponse<String> response = httpClient.send(
+                    request, HttpResponse.BodyHandlers.ofString());
+            ResponseDataDto responseDataDto = objectMapper.readValue(
```
olekskov: 1) Naming of class ResponseDataDto gives absolutely no information what is it
2) Why we need this additional DTO with List? Can we map directly to List<CharacterResponseDto>?
3) I don't like CharacterResponseDto naming also. I expect that this is something we produce outside of our api but this is model we obtain from external resource, maybe something like ExternalCharacter?

Then we get List<ExternalCharacter> which is rather understandable and easy to read

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Override
+    public void saveAll(List<CharacterResponseDto> listDto) {
+        List<Character> characters = listDto.stream()
+                .map(mapper::toModel)
+                .toList();
+        countCharacter = characters.size();
+        characterRepository.saveAll(characters);
+    }
```
olekskov: Cannot agree this logic should be here
You have character client which single task is get info from external resource into our DB.
It doesn't have to call CharacterService which is used for processing continious users requests

olekskov (changes requested):

kozhukhovsky (approved):

PR #36, opened 2023-11-02
Hometask completed

MarynaKushniruk:


src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+@Tag(name = "Character management", description = "Endpoints for getting characters from db")
+@RequiredArgsConstructor
+@RestController
+@RequestMapping(value = "/characters")
+public class CharacterController {
+    private final CharacterService characterService;
+
+    @GetMapping
```
olekskov: in rest usually GET method on /model endpoint without any arguments stands for getAll operation
I'd add something like /random to path to avoid confusion for user who hasn't read our swagger

src/main/java/mate/academy/rickandmorty/service/CharacterClientImpl.java
```diff
@@ -0,0 +1,40 @@
...
+    public List<CharacterResponseDto> getCharacters() {
+        HttpClient httpClient = HttpClient.newHttpClient();
+        HttpRequest request = HttpRequest.newBuilder()
+                .GET()
+                .uri(URI.create(CHARACTER_URL))
+                .build();
+        try {
+            HttpResponse<String> response =
+                    httpClient.send(request, HttpResponse.BodyHandlers.ofString());
+            ResponseDataDto responseDataDto =
```
olekskov: Naming of class ResponseDataDto gives absolutely no information what is it.
I don't like CharacterResponseDto naming also. 

In both cases, according to name, I expect that this is something we produce outside of our api but these are models we obtain from external resource

instead of  CharacterResponseDto maybe something like ExternalCharacter?
please also think of new name for ResponseDataDto

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+import mate.academy.rickandmorty.dto.internal.CharacterDto;
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import mate.academy.rickandmorty.model.Character;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import org.springframework.stereotype.Service;
+
+@RequiredArgsConstructor
+@Service
+public class CharacterServiceImpl implements CharacterService {
+    private static int countCharacters;
```
olekskov: redundant class variable, should be removed

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+    private static final Random RANDOM = new Random();
+    private final CharacterRepository characterRepository;
+    private final CharacterMapper characterMapper;
+
+    @Override
+    public CharacterDto getRandomCharacter() {
+        int id = RANDOM.nextInt(countCharacters) + 1;
```
olekskov: use nextLong() method and avoid casting int to long

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    public void saveAll(List<CharacterResponseDto> listDto) {
+        List<Character> characters = listDto.stream()
+                .map(characterMapper::toModel)
+                .toList();
+        countCharacters = characters.size();
+        characterRepository.saveAll(characters);
+
+    }
```
olekskov: Cannot agree this logic should be here
You have character client which single task is get info from external resource into our DB.
It doesn't have to call CharacterService which is used for processing continious users requests

olekskov (changes requested):

kozhukhovsky (approved):

PR #37, opened 2023-11-02
Implemented the project according to the requirements of the task, bu…jjj

0-zagar-0:
…t made some methods so that they can be used for other data from the Rick and Morty service


Ivan95kos (changes requested):
Please, fix all checks
Знімок екрана 2023-11-03 о 16 49 34

src/main/java/mate/academy/rickandmorty/service/ResourceService.java
```diff
@@ -0,0 +1,11 @@
+package mate.academy.rickandmorty.service;
+
+public interface ResourceService<R, C, S> {
```
Rommelua: Why do we need this interface?
image

Rommelua (changes requested):

src/main/java/mate/academy/rickandmorty/service/impl/ClientServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import lombok.RequiredArgsConstructor;
+import mate.academy.rickandmorty.service.ClientService;
+import mate.academy.rickandmorty.service.UrlService;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+// R - resource data page
```
kozhukhovsky, line 16:
```suggestion
```

kozhukhovsky (approved):

PR #39, opened 2023-11-03
Implemented according to the requirements of the task, but made some methods more abstract so that they can be applied to other links on the service

0-zagar-0:


0-zagar-0:
Implemented according to the requirements of the task, but made some methods more abstract so that they can be applied to other links on the service

created a new branch because something was wrong with that one

PR #44, opened 2023-11-05
Added working solution

IvanFadieiev:


src/main/java/mate/academy/rickandmorty/client/CharacterClient.java
```diff
@@ -0,0 +1,73 @@
...
+                    .build();
+            HttpResponse<String> firstResponse = httpClient.send(initialRequest,
+                    HttpResponse.BodyHandlers.ofString());
+            Long pages = objectMapper.readValue(objectMapper
+                    .readTree(firstResponse.body())
+                    .get("info")
+                    .get("pages")
+                    .toString(), Long.class);
+            for (int i = 1; i <= pages; i++) {
+                HttpRequest request = HttpRequest.newBuilder()
```
andrii-hoienko, line 43: code duplicatioon

src/main/java/mate/academy/rickandmorty/client/CharacterClient.java
```diff
@@ -0,0 +1,73 @@
...
+@Service
+@RequiredArgsConstructor
+public class CharacterClient {
+    private static final String BASE_URL = "https://rickandmortyapi.com/api/character";
+    private static final HttpClient httpClient = HttpClient.newHttpClient();
+    private final ObjectMapper objectMapper;
+    private final CharacterMapper characterMapper;
+    private final CharacterRepository characterRepository;
+
+    public void fetchDataFromApi() {
```
andrii-hoienko, line 29: too much responsibilities for one method, I suggest to introduce method (maybe class) which will be responsible for making request (for example lets name it makeRequest(url) and another one which responsibility will be to call makeRequest and collect responses. Maybe makes sence to create method readInitialResponse and readCharacterResponse for parsing raw data to objects

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImplementation.java
```diff
@@ -0,0 +1,35 @@
...
+    private final CharacterRepository characterRepository;
+    private final CharacterMapper characterMapper;
+
+    @Override
+    public CharacterResponseDto getRandomCharacter() {
+        Random random = new Random();
+        return characterMapper.toDto(characterRepository
+                .findById(random.nextLong(characterRepository
+                        .findAll().size())).orElseThrow(()
+                                -> new RuntimeException("Entity was not found")));
```
andrii-hoienko, line 25: It is bad practice to throw general exception, better to throw NotFoundException or something like this

andrii-hoienko (changes requested):
GJ! But some refactoring required

PR #58, opened 2023-11-13
add repository, service, mapper, controller & third-party API client …

VoitseshukV:
…implementation


src/main/java/mate/academy/rickandmorty/Application.java
```diff
@@ -1,12 +1,32 @@
...
+    @Component
+    @RequiredArgsConstructor
+    public class DataLoader implements ApplicationRunner {
+        private final CartoonCharacterClient characterClient;
+        private final CartoonCharacterService characterService;
+
+        public void run(ApplicationArguments args) {
+            List<CreateCartoonCharacterRequestDto> characters = characterClient.getAllCharacters();
+            characterService.saveAll(characters);
+        }
```
Sanderfuck: Why do you use Component for this and create inner class?

src/main/java/mate/academy/rickandmorty/controller/CartoonCharacterController.java
```diff
@@ -0,0 +1,35 @@
...
+@RestController
+@RequestMapping(value = "/api/characters")
+public class CartoonCharacterController {
+    private final CartoonCharacterService cartoonCharacterService;
+
+    @Operation(summary = "Get random character",
+            description = "Get random character from DB")
+    @GetMapping(value = "/random")
+    public CartoonCharacterDto getByName() {
+        return cartoonCharacterService.getRandom();
```
Sanderfuck: Is it correct naming of method?

src/main/java/mate/academy/rickandmorty/service/impl/CartoonCharacterClientImpl.java
```diff
@@ -0,0 +1,55 @@
...
+                } else {
+                    result.addAll(cartoonCharactersListResponseDto.results());
+                }
+            }
+            if (result == null || result.isEmpty()) {
+                throw new GettingCharactersListException("Error getting characters list");
+            }
+            return result;
+        } catch (IOException | InterruptedException e) {
+            throw new GettingCharactersListException("Error getting characters list", e);
```
Sanderfuck: Can you divide logic for one helpful method?

src/main/java/mate/academy/rickandmorty/service/impl/CartoonCharacterServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Override
+    public List<CartoonCharacterDto> findByName(String name, Pageable pageable) {
+        List<CartoonCharacter> characters = repository.findAllByNameContains(pageable, name);
+        if (characters.isEmpty()) {
+            throw new EntityNotFoundException("Can't find characters with name: " + name);
+        }
+        return characters.stream()
+                .map(mapper::toDto)
+                .toList();
```
Sanderfuck: This logic better place to service layer

Sanderfuck (changes requested):

src/main/java/mate/academy/rickandmorty/Application.java
VoitseshukV: I need to get service layer beans. Can't get them directly in a static main method. The option with initialization in ApplicationRunner was considered in one of the lessons. What other options are there?

VoitseshukV: I like the new implementation because of its compactness, but I am not sure that it is correct. I added only a fully encapsulated ThirdPartyApiDataLoader class bean to the Application. After its instance is created, its PostConstruct method is called, which performs data loading.

olekskov (approved):
Good job

PR #70, opened 2023-12-04
solution

kateryna-mykh:


okuzan (changes requested):
PR requested by mentor for debugging

src/main/java/mate/academy/rickandmorty/service/CharacterClient.java
```diff
@@ -0,0 +1,62 @@
...
+        for (int i = 1; i <= countOfPages; i++) {
+            request = HttpRequest.newBuilder().GET().uri(URI.create(URL)).build();
+            HttpResponse<String> response;
+            try {
+                response = client.send(request, HttpResponse.BodyHandlers.ofString());
+                CharacterResponseDto responseDto = objectMapper.readValue(response.body(),
+                        CharacterResponseDto.class);
+                
+                characters.addAll(responseDto.results());
+                URL = responseDto.info().next();
```
koval-kn: if you modify static value dynamically, it can cause errors. consider changing `URL` to non-static field or create extra method-level variable for that purpose

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import org.springframework.stereotype.Service;
+
+@RequiredArgsConstructor
+@Service
+public class CharacterServiceImpl implements CharacterService {
+    private final CharacterRepository repository;
+    private final CharacterMapper mapper;
+    private final Long maxId = 257L;
```
koval-kn:
```suggestion
```

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public CharacterDto getRandomCharacter() {
+        long generatedRandomId = new Random().nextLong(1, maxId);
```
koval-kn: 1. what if we'd have more than 257 characters? 
2. random should be `private final` class-level field
```suggestion
        long size = characterRepository.count();
        long generatedRandomId = random.nextLong(1, size);
```

koval-kn (changes requested):

src/main/java/mate/academy/rickandmorty/service/CharacterClient.java
kateryna-mykh: Ok

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
kateryna-mykh: @kerrrusha, 
1) if I count size in getRandomCharacter method it would be requested DB 2 times per every request. Is it ok? First I wrote this in case data would be modified/deleted:
` private Long getMaxId() {
        return repository.findAll().stream()
                .map(CustomCharacter::getId)
                .sorted(Comparator.reverseOrder())
                .findFirst()
                .get();
    }`   
2) Ok.

kozhukhovsky (approved):

PR #75, opened 2023-12-14
Created API with ability to get random character and find by name

dima-semeniuk:


src/main/java/mate/academy/rickandmorty/client/CharacterClient.java
```diff
@@ -0,0 +1,71 @@
...
+                dtos.addAll(resultDto.getResults());
+                page++;
+            }
+        } catch (IOException | InterruptedException e) {
+            throw new RuntimeException("Characters not found.", e);
+        }
+        return dtos;
+    }
+
+    public CharacterInfoDto getCharacterInfo() {
```
vovasalyha: You don't need to fetch the last page to know the count. Simply iterate through pages and use field `next` from the response. When you see `next` field is null it means you reached the last page.

vovasalyha (approved):
Looks good, but review your logic of iterating through the pages. You make one redundant API call to know the pages count, which is not optimal.

PR #76, opened 2023-12-15
Created app that works with RickAndMortyApi

HovorukhaBohdan:
Implemented `getWikiAboutRandomCharacter()` and `getCharactersByName()` methods


README.md
```diff
@@ -4,7 +4,7 @@
 
 - Task: Create API. It should contain two methods:
 
-1. The request randomly generates a wiki about one character in the universe the animated series Rick & Morty.
+1. The request randomly generates a wiki about one characterFromRickAndMorty in the universe the animated series Rick & Morty.
```
vovasalyha: You are not supposed to edit README file. I guess it's a bug with IntelliJ IDEA global rename tools, but pay attention to what you are pushing to git.

src/main/java/mate/academy/rickandmorty/db/init/DatabaseContent.java
```diff
@@ -0,0 +1,27 @@
...
+import lombok.RequiredArgsConstructor;
+import mate.academy.rickandmorty.client.RickAndMortyApiClient;
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import mate.academy.rickandmorty.model.CharacterFromRickAndMorty;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import org.springframework.stereotype.Component;
+
+@RequiredArgsConstructor
+@Component
+public class DatabaseContent {
```
vovasalyha: IMO not the best class name. Use something like DatabaseInitService or similar.

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+@RequiredArgsConstructor
+@Service
+public class CharacterServiceImpl implements CharacterService {
+    private final CharacterRepository characterRepository;
+    private final CharacterMapper characterMapper;
+    private final Random random = new Random();
+
+    @Override
+    public CharacterDto getWikiAboutRandomCharacter() {
+        long numberOfCharacters = characterRepository.count();
```
vovasalyha: Do you really need to fetch characters count each time the method is called? Charaters are fetched from API only once, so you can easily cache it.

vovasalyha (changes requested):

kshuryhin (approved):
looks good

PR #78, opened 2023-12-20
"jv-rick-and-morty" solution

OlexVoit:
added structures project. created compulsory classes. Implemented by API Client through FeignClient

I have changed the pom.xml file because I got a little confused and needed to update the spring boot version. 
Spring boot was not compatible with FeignClient.


src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+        return characterRepository.findAllByNameIsLikeIgnoreCase(namePart).stream()
+                .map(characterMapper::toDto)
+                .toList();
+
+    }
+
+    @Override
+    public CharacterDto getRandomCharacter() {
+        CharacterFromRickAndMorty randomCharacter = characterRepository
+                .findById(random.nextLong(characterRepository.count()))
```
vovasalyha: Do you really need to fetch a count from DB every time method is called? Because characters are fetched only once during startup and it seems like a redundant call to DB. You can fetch it one time instead.

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import mate.academy.rickandmorty.model.CharacterFromRickAndMorty;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import mate.academy.rickandmorty.service.CharacterService;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class CharacterServiceImpl implements CharacterService {
+    private static final String NOY_FOUND_RANDOM_ENTITY = "Couldn't get a random entity";
```
vovasalyha:
```suggestion
    private static final String RANDOM_ENTITY_NOT_FOUND_MSG = "Couldn't get a random entity";
```

src/main/java/mate/academy/rickandmorty/service/impl/RickAndMortyClientToDbServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import mate.academy.rickandmorty.model.CharacterFromRickAndMorty;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import mate.academy.rickandmorty.service.RickAndMortyClientToDbService;
+import mate.academy.rickandmorty.service.rick.and.morty.client.RickAndMortyClient;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class RickAndMortyClientToDbServiceImpl implements RickAndMortyClientToDbService {
```
vovasalyha: From the class name we can already see that we break SRP principle. I would create a `getAllCharacters` method and call this service something like DbInitService or similar which simply calls client service and stores everything in DB. This way we don't violate SOLID.

src/main/java/mate/academy/rickandmorty/AppConfig.java
```diff
@@ -0,0 +1,13 @@
+package mate.academy.rickandmorty;
+
+import java.util.Random;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class AppConfig {
+    @Bean
+    public Random random() {
```
vovasalyha: I'm not sure that we need to create a bean for Random class. You can even use ThreadLocalRandom directly where you need it (it will give even better randomness in multithreaded environment).

vovasalyha (changes requested):

OlexVoit:
and AppConfig was also removed

README.md
```diff
@@ -39,3 +39,4 @@
   and `src/test/resources/application.properties` files. In other case you may face a problem with Application Context
   creation during the `mvn test` phase.
 - Requests must be documented using Swagger.
+- test
```
d1sam:
```suggestion
```

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    public CharacterDto getRandomCharacter() {
+        CharacterFromRickAndMorty randomCharacter = characterRepository
+                .findById(random.nextLong(getNumberOfCharactersInDb()))
+                .orElseThrow(() -> new EntityNotFoundException(RANDOM_ENTITY_NOT_FOUND_MSG));
+        return characterMapper.toDto(randomCharacter);
+    }
+
+    @CachePut(cacheNames = "numberOfCharacters")
+    public long getNumberOfCharactersInDb() {
```
d1sam: it's better to use `@PostConstruct` and set in the class field `numberOfCharacters`

d1sam (changes requested):
Some changes should be done:)

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    public CharacterDto getRandomCharacter() {
+        CharacterFromRickAndMorty randomCharacter = characterRepository
+                .findById(random.nextLong(getNumberOfCharactersInDb()))
+                .orElseThrow(() -> new EntityNotFoundException(RANDOM_ENTITY_NOT_FOUND_MSG));
+        return characterMapper.toDto(randomCharacter);
+    }
+
+    @PostConstruct
+    public long getNumberOfCharactersInDb() {
+        return characterRepository.count();
```
d1sam, line 41: you should set `characterRepository.count();` in the class field numberOfCharacters and then use this value in `getRandomCharacter()` method. don't call `getNumberOfCharactersInDb()` method in `getRandomCharacter()`

d1sam (changes requested):
Good job, but one minor change should be done!

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Override
+    public CharacterDto getRandomCharacter() {
+        CharacterFromRickAndMorty randomCharacter = characterRepository
+                .findById(random.nextLong(getNumberOfCharactersInDb()))
+                .orElseThrow(() -> new EntityNotFoundException(RANDOM_ENTITY_NOT_FOUND_MSG));
+        return characterMapper.toDto(randomCharacter);
+    }
+
+    @Cacheable(cacheNames = "numberOfCharacters")
```
vovasalyha, line 39: Cacheable is good, but looks a bit overcomplicated for such trivial task. (But for learning purposes it's ok)

vovasalyha (approved):

PR #79, opened 2023-12-20
Roma Menzheres's realization of "Rick and Morty" task

RomanMenzheres:
Created api that have two endpoints: 1) Get wiki about random character 2) Find characters by name

### DIFF
--- a/src/main/java/mate/academy/rickandmorty/Application.java
+++ b/src/main/java/mate/academy/rickandmorty/Application.java
@@ -1,12 +1,25 @@
 package mate.academy.rickandmorty;
 
+import mate.academy.rickandmorty.client.SeriesCharacterClient;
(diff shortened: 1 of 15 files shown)


src/main/java/mate/academy/rickandmorty/Application.java
```diff
@@ -1,12 +1,25 @@
...
+import org.springframework.context.annotation.Bean;
 
 @SpringBootApplication
 public class Application {
 
     public static void main(String[] args) {
         SpringApplication.run(Application.class, args);
     }
+
+    @Bean
```
Ivan95kos, line 16: For the future, you can use the annotation `@PostConstract` instead of it

Ivan95kos (approved):
Nice

PR #84, opened 2024-01-01
Java Rick And Morty First Iteration

deikunx:


olekskov (approved):
Good job. 
Like that you fully separated service that works with your controller and service that is responsible to take data from external api.
Also nice work on repository methods.

PR #89, opened 2024-01-11
Hw rickmorty

DZarudnyi:


src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+
+@Service
+@RequiredArgsConstructor
+public class CharacterServiceImpl implements CharacterService {
+    private final CharacterRepository characterRepository;
+    private final CharacterMapper mapper;
+
+    @Override
+    public CharacterRequestDto getRandomCharacter() {
+        Random random = new Random();
```
okuzan: make it a class-level field

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiClient.java
```diff
@@ -0,0 +1,37 @@
...
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import lombok.RequiredArgsConstructor;
+import mate.academy.rickandmorty.dto.external.CharacterResponseDataDto;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class RickAndMortyApiClient {
+
```
okuzan:
```suggestion
```

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiClient.java
```diff
@@ -0,0 +1,37 @@
...
+    public CharacterResponseDataDto getCharactersInfo(String url) {
+        HttpClient httpClient = HttpClient.newHttpClient();
+        HttpRequest httpRequest = HttpRequest.newBuilder()
+                .GET()
+                .uri(URI.create(url))
+                .build();
+        try {
+            HttpResponse<String> response =
+                    httpClient.send(httpRequest, HttpResponse.BodyHandlers.ofString());
+            System.out.println(response.body());
```
okuzan: don't forget to remove such things before committing

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiClient.java
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class RickAndMortyApiClient {
+
+    private final ObjectMapper objectMapper;
+
+    public CharacterResponseDataDto getCharactersInfo(String url) {
+        HttpClient httpClient = HttpClient.newHttpClient();
```
okuzan: make it a field and reuse it

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+        return characterRepository.findAllByNameContainsIgnoreCase(name).stream()
+                .map(mapper::toDto)
+                .toList();
+    }
+
+    @Override
+    public void saveAll(List<CharacterResponseDto> characterResponseDtos) {
+        for (CharacterResponseDto character : characterResponseDtos) {
+            characterRepository.save(mapper.toModel(character));
+        }
```
okuzan: let's optimize this by using `saveAll`
```suggestion
    List<Character> characters = characterResponseDtos.stream()
                                                      .map(mapper::toModel)
                                                      .collect(Collectors.toList());
    characterRepository.saveAll(characters);
```

okuzan (changes requested):
Good job, clean code

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
DZarudnyi: Tried to do this, but something went wrong, i think i didn't convert it to model before saving. Thanks for pointing that out fot me

DZarudnyi: Oh, and one more thing. Will it be wrong, if i won't declare a variable and do all the operations inside the parameters of saveAll() operation?

olekskov: would be compiled for sure but
1) Not really the best structure to read
```
    characterRepository.saveAll(characterResponseDtos.stream()
                                                      .map(mapper::toModel)
                                                      .toList());
```
2) Harder to debug. You actually cannot place any valuable breakpoint so you will have to work with evaluate expression only

But this is my personal opinion. Don't like 1-liner methods due to more effort required to work with them

olekskov (approved):
Good job

PR #90, opened 2024-01-12
Rick And Morty API:

GoldHolub:
- added functionality to find Character or Characters by name and get random character
- added documentation by using Swagger
- used MySQL and H2 DB in tests


olekskov (approved):
Good job

PR #100, opened 2024-01-26
add solution

slaybrute:


olekskov (changes requested):
Major mismatch of solution with requirements:
1) "All data from the public API should be fetched once, and only once, when the Application context is created"
but you solution passes each request to external api
2) "Use MySQL DB in your app."
cannot see any DB used

Ivan95kos (changes requested):
"All data from the public API should be fetched once, and only once, when the Application context is created"
but you solution passes each request to external api

slaybrute:
I've added to internal as well check my last commit please

Ivan95kos (approved):
Nice!

PR #102, opened 2024-01-31
added solution for this task

polikarpov13:


src/main/resources/application.properties
```diff
@@ -1 +1,5 @@
-
+spring.datasource.url=jdbc:h2:mem:testdb
```
olekskov: According to task requirements: Use MySQL DB in your app

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class RickAndMortyApiServiceImpl implements RickAndMortyApiService {
+    private final CharacterRepository characterRepository;
+
+    @Override
+    public Character getRandomCharacter() {
```
olekskov: I'd recommend using DTO even for the simplest cases instead of passing Entity class outside.
You never know how you will need to change your DB structure tomorrow and it's not good to change your API together with it

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiDataServiceImpl.java
```diff
@@ -0,0 +1,60 @@
...
+                .GET()
+                .uri(URI.create(URL))
+                .build();
+        try {
+            HttpResponse<String> response = httpClient.send(httpRequest,
+                    HttpResponse.BodyHandlers.ofString());
+
+            CharacterResultResponseDto responseDto = objectMapper.readValue(
+                    response.body(), CharacterResultResponseDto.class);
+            List<Character> characters = convertCharacterApiDtosToModelList(responseDto);
```
olekskov: Please pay attention that external api has mandatory pagination. You cannot get all characters within one call, now you get only first page

olekskov (changes requested):
nice solution overall, but there are some comments

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiDataServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+        try {
+            return httpClient.send(httpRequest, HttpResponse.BodyHandlers.ofString());
+        } catch (IOException | InterruptedException e) {
+            throw new RuntimeException("Could not send reply back!", e);
+        }
+    }
+
+    private List<Character> convertCharacterApiDtosToModelList(CharacterResultResponseDto dtos) {
+        return dtos.results().stream()
+                .map(this::toModel)
```
okuzan, line 60:
```suggestion
        return dtos.results()
                .stream()
                .map(this::toModel)

```

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiDataServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+    private static final String URL = "https://rickandmortyapi.com/api/character";
+    private static final int PAGE_AMOUNT = 42;
+    private final CharacterRepository characterRepository;
+    private final CharacterMapper characterMapper;
+    private final ObjectMapper objectMapper;
+
+    @Override
+    public void fetchDataAndSaveToDatabase() {
+        try {
+            for (int i = 1; i <= PAGE_AMOUNT; i++) {
```
okuzan, line 30:
```suggestion
            for (int i = 0; i < PAGE_AMOUNT; i++) {
```

okuzan (approved):
Well done

PR #112, opened 2024-02-12
Add RickAndMortyApi support, Endpoints for work with custom api

IhorTrokhymchuk:


src/main/java/mate/academy/rickandmorty/exception/CustomGlobalExceptionHandler.java
```diff
@@ -0,0 +1,62 @@
...
+        errorsBody.put("timestamp", LocalDateTime.now());
+        errorsBody.put("status", HttpStatus.BAD_REQUEST);
+        final List<String> messagesList = ex.getBindingResult().getAllErrors().stream()
+                .map(this::getErrorMessage)
+                .toList();
+        errorsBody.put("errors", messagesList);
+        return new ResponseEntity<>(errorsBody, headers, status);
+    }
+
+    private String getErrorMessage(ObjectError e) {
```
olekskov, line 45: private methods should be placed after public ones

src/main/java/mate/academy/rickandmorty/controller/CharactersController.java
```diff
@@ -0,0 +1,34 @@
...
+    private final CharacterService characterService;
+
+    @GetMapping("/random")
+    @Operation(summary = "Get random Character",
+            description = "Get one random Character from DB")
+    public CharacterDto getAll() {
+        return characterService.getRandom();
+    }
+
+    @GetMapping("/by_name")
```
olekskov: I'd remove url suffix.
Filtered search is like a prolongation of getAll method that should be available without suffix.
Agree, that it can be a bit confusing because you don't have getAll method. But getAll is not required by the task.
And I don't think this is the reason to change endpoints naming convention
```suggestion
    @GetMapping
```

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+    public void insertCharactersToDB() {
+        if (startPage < 1 || endPage < 1 || startPage > endPage) {
+            throw new StartApplicationException("Dont valid start or end pages in properties");
+        }
+        for (int i = startPage; i <= endPage; i++) {
+            characterRepository.saveAll(fetchDataToModel(
+                    fetchApiService.fetchDataFromApi(String.format(url, i))
+            ));
+        }
+    }
```
olekskov: Wish to see this part in a separate component (with related private methods).
You have variables and logic that are designed for one specific purpose and isn't similar with other public methods in this service

olekskov (changes requested):

okuzan (approved):

PR #115, opened 2024-02-14
Implemented getRandom and getAllByNamePart methods

nshtykh:


src/main/java/mate/academy/rickandmorty/config/AppConfig.java
```diff
@@ -0,0 +1,13 @@
+package mate.academy.rickandmorty.config;
+
+import java.util.Random;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class AppConfig {
+    @Bean
+    public Random random() {
```
vovasalyha: You don't need a bean of `Random` class. Also, in concurrent environments (web app is concurrent) it's preferred to use `ThreadLocalRandom`

src/main/java/mate/academy/rickandmorty/controller/CharactersController.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/characters")
+@RequiredArgsConstructor
+@Tag(name = "Characters management",
+        description = "Endpoints for managing characters from the Rick and Morty cartoon")
```
vovasalyha: Usually `managing` means that the API allows some creation/modification, but this API is only for read access.
```suggestion
        description = "Endpoints for retrieving characters from the Rick and Morty cartoon")
```

src/main/java/mate/academy/rickandmorty/controller/CharactersController.java
```diff
@@ -0,0 +1,34 @@
...
+    private final CharacterService characterService;
+
+    @GetMapping("/random")
+    @Operation(summary = "Get random character")
+    public CharacterInternalDto getRandom() {
+        return characterService.getRandomCharacter();
+    }
+
+    @GetMapping("/by-name")
+    @Operation(summary = "Get all characters by name part")
```
vovasalyha: I would also specify that this search ignores case.

src/main/java/mate/academy/rickandmorty/service/character/CharacterServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    private final CharacterMapper mapper;
+
+    @Override
+    public void saveAll(List<Character> characters) {
+        characterRepository.saveAll(characters);
+    }
+
+    @Override
+    public CharacterInternalDto getRandomCharacter() {
+        long randomId = random.nextLong(characterRepository.count());
```
vovasalyha: Think about optimizing this logic (considering that you save all characters only once on application startup)

vovasalyha (changes requested):

vovasalyha (approved):

PR #116, opened 2024-02-15
RIck & Morty task implemented

MateuszRostek:


src/main/resources/application.properties
nowickimj, line 1: On of the requirements goes:
```
Keep identical set of params in the src/main/resources/application.properties and src/test/resources/application.properties files. In other case you may face a problem with Application Context creation during the mvn test phase.
```

src/test/java/mate/academy/rickandmorty/ApplicationTests.java
```diff
@@ -1,13 +0,0 @@
-package mate.academy.rickandmorty;
-
-import org.junit.jupiter.api.Test;
-import org.springframework.boot.test.context.SpringBootTest;
-
-@SpringBootTest
-class ApplicationTests {
```
nowickimj, line 7: Why did you delete this class?

src/main/java/mate/academy/rickandmorty/service/RickAndMortyServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+import mate.academy.rickandmorty.mapper.RickAndMortyMapper;
+import mate.academy.rickandmorty.repository.RickAndMortyRepository;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class RickAndMortyServiceImpl implements RickAndMortyService {
+    private final RickAndMortyRepository rickAndMortyRepository;
+    private final RickAndMortyMapper rickAndMortyMapper;
+    private Long totalAmountOfCharacters;
```
nowickimj: In real world scenarios, it's a bad practice to store non-constant values within the Spring components as Spring does not provide out-of-the-box thread safety for beans.
There is no reason for now to cache this value, you may just call `count()` every time a random character is being fetched.

src/main/java/mate/academy/rickandmorty/service/RickAndMortyServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public RickAndMortyCharacterDto getRandomCharacter() {
+        if (totalAmountOfCharacters == null) {
+            totalAmountOfCharacters = rickAndMortyRepository.count();
+        }
+        Long randomId = new Random().nextLong(totalAmountOfCharacters) + 1;
```
nowickimj: `Random` object be a constant

src/main/java/mate/academy/rickandmorty/service/RickAndMortyServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+        return rickAndMortyMapper.toDto(
+                rickAndMortyRepository.findById(randomId)
+                        .orElseThrow(
```
nowickimj: Redundant new lines here - the whole `.orElseThrow()` implementation is clear enough for single line:
```java
var randomCharacter = rickAndMortyRepository.findById(randomId)
  .orElseThrow(() -> new RuntimeException("Can't get a random character from db"));
return rickAndMortyMapper.toDto(randomCharacter);
```

src/main/java/mate/academy/rickandmorty/service/RickAndMortyServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+                                () -> new RuntimeException(
+                                        "Can't get a random character from db")));
+    }
+
+    @Override
+    public List<RickAndMortyCharacterDto> findCharactersByNameContaining(String nameSegment) {
+        return rickAndMortyRepository.findAllByNameContainsIgnoreCase(nameSegment).stream()
+                .map(rickAndMortyMapper::toDto).toList();
```
nowickimj: It's very common for streams implementation to put every method in the chain, like this:
```java
someList.stream()
  .filter(Objects::nonNull)
  .map(Object::toString)
  .toList();
```

src/main/java/mate/academy/rickandmorty/service/DataInitializationRickAndMortyService.java
```diff
@@ -0,0 +1,22 @@
...
+
+@Service
+@RequiredArgsConstructor
+public class DataInitializationRickAndMortyService implements DataInitializationService {
+    private final RickAndMortyApiClient apiClient;
+    private final RickAndMortyMapper rickAndMortyMapper;
+    private final RickAndMortyRepository apiRepository;
+
+    @Override
+    public void initializeDbData() {
```
nowickimj, line 17: You can replace the interface usage with `@PostConstruct` annotation for method that will be called once after the component initialization, for example:
```java
@PostConstruct
public void init() {
  apiRepository.saveAll(/* your code goes here */);
}
```

- delete unnecessary interface
- remove runner initialization and injection from main class

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiClient.java
```diff
@@ -0,0 +1,49 @@
...
+    private static final String BASE_URL = "https://rickandmortyapi.com/api/character/?page=";
+    private final ObjectMapper objectMapper;
+    @Value("${rick.and.morty.characters.last.page}")
+    private int lastPage;
+
+    public List<CharacterResultDto> fetchAllData() {
+        List<CharacterResultDto> allCharacter = new ArrayList<>();
+        HttpClient httpClient = HttpClient.newHttpClient();
+        int currentPage = 1;
+        while (currentPage <= lastPage) {
```
nowickimj: As you iterate over integers, please replace the `while` loop with the old good `for` - it's easier to read and more immune to mistakes.

src/main/java/mate/academy/rickandmorty/controller/RickAndMortyController.java
```diff
@@ -0,0 +1,43 @@
...
+import mate.academy.rickandmorty.dto.internal.RickAndMortyCharacterDto;
+import mate.academy.rickandmorty.service.RickAndMortyServiceImpl;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+@Tag(
+        name = "Rick & Morty Characters Information Service",
+        description = "Service for retrieving information about "
```
nowickimj: For long strings, you may use string blocks instead of concatenation:
```java
        description = """
            Service for retrieving information about
            characters in the Rick & Morty universe
        """)
```

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiClient.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class RickAndMortyApiClient {
+    private static final String BASE_URL = "https://rickandmortyapi.com/api/character/?page=";
+    private final ObjectMapper objectMapper;
+    @Value("${rick.and.morty.characters.last.page}")
+    private int lastPage;
```
Rommelua: we can not have it in properties - last page can be changed by external api owners

src/main/java/mate/academy/rickandmorty/dto/external/CharacterResponseDataDto.java
```diff
@@ -0,0 +1,9 @@
+package mate.academy.rickandmorty.dto.external;
+
+import java.util.List;
+import lombok.Data;
+
+@Data
```
nowickimj: Lombok's `@Data` produces lots of code underneath - most of it is redundant.
For DTOs, please stick with `record` type.

src/main/java/mate/academy/rickandmorty/service/RickAndMortyServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+    private final RickAndMortyMapper rickAndMortyMapper;
+    private Long totalAmountOfCharacters;
+
+    @Override
+    public RickAndMortyCharacterDto getRandomCharacter() {
+        if (totalAmountOfCharacters == null) {
+            totalAmountOfCharacters = rickAndMortyRepository.count();
+        }
+        Long randomId = new Random().nextLong(totalAmountOfCharacters) + 1;
+        return rickAndMortyMapper.toDto(
```
Rommelua: investigate how to fetch random object directly from DB

Rommelua (changes requested):

src/main/java/mate/academy/rickandmorty/dto/external/CharacterResultDto.java
```diff
@@ -0,0 +1,15 @@
+package mate.academy.rickandmorty.dto.external;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Data;
+
+@Data
+public class CharacterResultDto {
+    @JsonProperty("id")
```
nowickimj: Why not just call this field `id`?

src/main/java/mate/academy/rickandmorty/dto/external/CharacterResultDto.java
```diff
@@ -0,0 +1,15 @@
+package mate.academy.rickandmorty.dto.external;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Data;
+
+@Data
+public class CharacterResultDto {
```
nowickimj: likewise, use record

src/main/java/mate/academy/rickandmorty/dto/internal/RickAndMortyCharacterDto.java
```diff
@@ -0,0 +1,14 @@
+package mate.academy.rickandmorty.dto.internal;
+
+import lombok.Data;
+
+@Data
+public class RickAndMortyCharacterDto {
```
nowickimj: likewise, use record

nowickimj (changes requested):

src/test/java/mate/academy/rickandmorty/ApplicationTests.java
MateuszRostek: I had some weird problems with the test while running `mvn clean package`. I tried to keep the same set of parameters, but after many attempts, I always ended up with an error, so I decided to delete it.

src/main/java/mate/academy/rickandmorty/service/RickAndMortyServiceImpl.java
MateuszRostek: I did it like this because in a single line it is longer than 100 characters, so I'm getting a checkstyle error. But now I changed the styling a little bit.

src/main/java/mate/academy/rickandmorty/dto/external/CharacterResultDto.java
MateuszRostek: But wouldn't I have to add the appropriate annotation to the "toModel" method from the mapper afterwards? I thought this way was more readable

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiClient.java
MateuszRostek: I had to remove info about last field from properties, so now i'm not sure about my loop implementation

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiClient.java
```diff
@@ -0,0 +1,50 @@
...
+public class RickAndMortyApiClient {
+    private static final String BASE_URL = "https://rickandmortyapi.com/api/character/?page=";
+    private final ObjectMapper objectMapper;
+
+    public List<CharacterResultDto> fetchAllData() {
+        List<CharacterResultDto> allCharacter = new ArrayList<>();
+        HttpClient httpClient = HttpClient.newHttpClient();
+        int currentPage = 1;
+        int lastPage = 1;
+        for (int i = currentPage; i <= lastPage; i++) {
```
nowickimj: Okay - @Rommelua is correct about the dynamic `lastPage` value, just checked the external API - in that case, using `for` is even worse idea as right now it's hard to tell what is going on.

How about slightly different approach: use `info.next` instead of `info.pages` with `while` loop
1. Create `requestUrl` variable with initial value: `https://rickandmortyapi.com/api/character`
2. In loop, until `requestUrl` is null:
  1.1. Call API
  1.2. Process respone
  1.3. Update `requestUrl` with value from response (on final iteration, its value will be null)

src/main/java/mate/academy/rickandmorty/service/RickAndMortyServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+import mate.academy.rickandmorty.mapper.RickAndMortyMapper;
+import mate.academy.rickandmorty.repository.RickAndMortyRepository;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class RickAndMortyServiceImpl implements RickAndMortyService {
+    private final RickAndMortyRepository rickAndMortyRepository;
+    private final RickAndMortyMapper rickAndMortyMapper;
+    private final Random random = new Random();
```
nowickimj: `private static final`

src/main/java/mate/academy/rickandmorty/dto/external/CharacterResultDto.java
```diff
@@ -0,0 +1,13 @@
+package mate.academy.rickandmorty.dto.external;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public record CharacterResultDto(
+        @JsonProperty("id")
```
nowickimj: I still don't understand the reason behind the usage of this annotation - properties in this class should be named as they are, as the class represents the entity in external system. Delete this annotation and rename property to `id`.

src/main/java/mate/academy/rickandmorty/Application.java
```diff
@@ -1,11 +1,12 @@
 package mate.academy.rickandmorty;
 
+import lombok.RequiredArgsConstructor;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
 @SpringBootApplication
+@RequiredArgsConstructor
```
nowickimj: No longer needed.

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiClient.java
```diff
@@ -0,0 +1,50 @@
...
+
+@Component
+@RequiredArgsConstructor
+public class RickAndMortyApiClient {
+    private static final String BASE_URL = "https://rickandmortyapi.com/api/character/?page=";
+    private final ObjectMapper objectMapper;
+
+    public List<CharacterResultDto> fetchAllData() {
+        List<CharacterResultDto> allCharacter = new ArrayList<>();
+        HttpClient httpClient = HttpClient.newHttpClient();
```
nowickimj: HttpClient instance can be a constant (private static final)

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiClient.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class RickAndMortyApiClient {
+    private static final String BASE_URL = "https://rickandmortyapi.com/api/character/?page=";
+    private final ObjectMapper objectMapper;
+
+    public List<CharacterResultDto> fetchAllData() {
+        List<CharacterResultDto> allCharacter = new ArrayList<>();
```
nowickimj: (minor) `allCharacters`

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApiClient.java
```diff
@@ -0,0 +1,50 @@
...
+import mate.academy.rickandmorty.dto.external.CharacterResultDto;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class RickAndMortyApiClient {
+    private static final String BASE_URL = "https://rickandmortyapi.com/api/character/?page=";
+    private final ObjectMapper objectMapper;
+
+    public List<CharacterResultDto> fetchAllData() {
```
nowickimj: `getAllCharacters` would make a better name - otherwise, the method name doesn't tell what data are you actually fetching, especially if more methods are created in future.

nowickimj (changes requested):

src/test/java/mate/academy/rickandmorty/ApplicationTests.java
nowickimj, line 1: I've tested your locally with this class present - everything works fine. 
I also see that code checks passed before the deletion.

Maybe you've dealt with a problem during the development that no longer occurs or there is a problem with your local maven.

This class shouldn't be deleted as you might want to provide unit tests in upcoming rounds - please revert it.

src/main/java/mate/academy/rickandmorty/Application.java
```diff
@@ -5,7 +5,6 @@
 
 @SpringBootApplication
 public class Application {
-
```
nowickimj: Unnecessary change - you don't need to apply changes that are not related to tasks.

nowickimj (changes requested):
minor 2 comments - please fix them and it's finished :)

MateuszRostek:
Thanks for checking the task so thoroughly, I got a lot out of it 😃

nowickimj (approved):

PR #118, opened 2024-02-16
First solution of jv-rick-and-morty

QbaSekowski:


nowickimj (changes requested):
Your code doesn't compile successfully - fix it before requesting a review.

src/main/java/mate/academy/rickandmorty/service/RickAndMortyClient.java
```diff
@@ -0,0 +1,38 @@
...
+                .uri(URI.create(BASE_URL))
+                .build();
+
+        try {
+            HttpResponse<String> response = httpClient.send(
+                    request, HttpResponse.BodyHandlers.ofString());
+
+            return objectMapper.readValue(
+                    response.body(), CharacterResponseDataDto.class);
+
```
okuzan: are you sure you receive all 826 characters with this implementation? Remember about pagination. You need some kind of a loop to check if there's a next page
image

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+    private final CharacterMapper characterMapper;
+    private final CharacterRepository characterRepository;
+    private final Random random = new Random();
+
+    @Override
+    public CharacterInternalResponseDto getRandomCharacter() {
+        long randomId = random.nextLong(characterRepository.count());
+        Character character = characterRepository.findById(randomId)
+                .orElseThrow(() ->
+                        new RuntimeException("Cannot find a character by id: " + randomId));
```
okuzan: it's better to throw more specific exceptions where possible, let's go for EntityNotFound or NoSuchElement exception here

src/main/java/mate/academy/rickandmorty/repository/CharacterRepository.java
```diff
@@ -0,0 +1,13 @@
...
+import java.util.List;
+import mate.academy.rickandmorty.model.Character;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface CharacterRepository extends JpaRepository<Character, Long> {
+    @Query(value = "SELECT c FROM Character c WHERE c.name LIKE CONCAT('%', :name, '%')")
+    List<Character> findAllByNameContainingIgnoreCase(String name);
```
okuzan: Consider removing the @Query annotation for this method as Spring Data JPA's query derivation from method names can automatically generate the required query

okuzan (changes requested):
Good job so far! Make sure to retrieve all the characters, keeping pagination in mind

src/main/java/mate/academy/rickandmorty/Application.java
```diff
@@ -5,7 +5,6 @@
 
 @SpringBootApplication
 public class Application {
-
```
nowickimj: Avoid unnecessary changes in your PRs.

src/main/java/mate/academy/rickandmorty/dto/external/CharacterExternalResponseDto.java
```diff
@@ -0,0 +1,21 @@
+package mate.academy.rickandmorty.dto.external;
+
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import lombok.Data;
+
+@Data
+public class CharacterExternalResponseDto {
```
nowickimj: You might use record instead of regular class for DTOs. Additionally, you won't require Lombok - `@Data` produces lots of boilerplate code.

src/main/java/mate/academy/rickandmorty/dto/external/CharacterResponseDataDto.java
```diff
@@ -0,0 +1,10 @@
+package mate.academy.rickandmorty.dto.external;
+
+import java.util.List;
+import lombok.Data;
+
+@Data
+public class CharacterResponseDataDto {
```
nowickimj: Same here, you can use record.

src/test/java/mate/academy/rickandmorty/ApplicationTests.java
```diff
@@ -5,9 +5,7 @@
 
 @SpringBootTest
 class ApplicationTests {
-
-	@Test
-	void contextLoads() {
-	}
-
+    @Test
```
nowickimj, line 9: Unnecessary change.

nowickimj (changes requested):
Few minor comments, good work with the rest :)

nowickimj (approved):

PR #124, opened 2024-02-26
implemented connection to API; added CharacterController

DEz-1996:
added CharacterController

### DIFF
--- a/src/main/java/mate/academy/rickandmorty/service/DbFiller.java
+++ b/src/main/java/mate/academy/rickandmorty/service/DbFiller.java
@@ -1,0 +1,37 @@
+package mate.academy.rickandmorty.service;
+
+import jakarta.annotation.PostConstruct;
+import java.util.List;
+import lombok.RequiredArgsConstructor;
+import mate.academy.rickandmorty.dto.external.ApiResponseDto;
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import mate.academy.rickandmorty.model.Character;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import mate.academy.rickandmorty.service.imp.CharacterHttpClient;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class DbFiller {
+    private static final String URL = "https://rickandmortyapi.com/api/character?page=1";
+    private final CharacterMapper characterMapper;
+    private final CharacterHttpClient rickAndMortyClient;
+    private final CharacterRepository characterRepository;
+
+    @PostConstruct
+    public void loadDataToDb() {
+        String nextUrl = URL;
+        while (nextUrl != null) {
+            ApiResponseDto apiResponse = rickAndMortyClient.get(nextUrl, ApiResponseDto.class);
+            save(apiResponse);
+            nextUrl = apiResponse.info().next();
+        }
+    }
+
+    private void save(ApiResponseDto apiResponse) {
+        List<Character> characterList = apiResponse.data().stream()
+                .map(characterMapper::toEntity)
+                .toList();
+        characterRepository.saveAll(characterList);
+    }
+}
--- a/src/main/java/mate/academy/rickandmorty/service/imp/CharacterServiceImp.java
+++ b/src/main/java/mate/academy/rickandmorty/service/imp/CharacterServiceImp.java
@@ -1,0 +1,42 @@
+package mate.academy.rickandmorty.service.imp;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.Random;
+import lombok.RequiredArgsConstructor;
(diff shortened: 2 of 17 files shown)


src/main/java/mate/academy/rickandmorty/service/DbFiller.java
```diff
@@ -0,0 +1,37 @@
...
+    private final CharacterMapper characterMapper;
+    private final CharacterHttpClient rickAndMortyClient;
+    private final CharacterRepository characterRepository;
+
+    @PostConstruct
+    public void loadDataToDb() {
+        String nextUrl = URL;
+        while (nextUrl != null) {
+            ApiResponseDto apiResponse = rickAndMortyClient.get(nextUrl, ApiResponseDto.class);
+            save(apiResponse);
```
vovasalyha, line 26: It's better to store all results in a list and then call `saveAll` on all results.

src/main/java/mate/academy/rickandmorty/service/imp/CharacterServiceImp.java
```diff
@@ -0,0 +1,42 @@
...
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import mate.academy.rickandmorty.model.Character;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import mate.academy.rickandmorty.service.CharacterService;
+import org.springframework.data.domain.Pageable;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class CharacterServiceImp implements CharacterService {
```
vovasalyha, line 17:
```suggestion
public class CharacterServiceImpl implements CharacterService {
```

src/main/resources/application.properties
```diff
@@ -1 +1,13 @@
...
+spring.datasource.username=root
+spring.datasource.password=12345Qwert!
+spring.datasource.driver-class-name=com.mysql.cj.jdbc.Driver
 
+spring.jpa.properties.hibernate.dialect = org.hibernate.dialect.MySQL8Dialect
+spring.jpa.hibernate.ddl-auto=create-drop
+spring.jpa.show-sql=true
+spring.jpa.open-in-view=true
+
+spring.jackson.deserialization.fail-on-unknown-properties=true
```
vovasalyha, line 11: IMO it's a bad idea to fail on unknown properties. What if the API will change and add some new field? Then your app will probably stop working.

src/main/java/mate/academy/rickandmorty/service/imp/CharacterServiceImp.java
```diff
@@ -0,0 +1,42 @@
...
+    private static final String CANT_GET_RANDOM_MSG = "Can't get random character with id ";
+    private final CharacterRepository characterRepository;
+    private final CharacterMapper characterMapper;
+    private final Random random;
+
+    @Override
+    public CharacterResponseDto getRandomCharacter() {
+        long count = characterRepository.count();
```
vovasalyha, line 25: This `count` value can be easily cached, as out database is changed only on startup, after that the count will be the same.

src/main/java/mate/academy/rickandmorty/service/DbFiller.java
```diff
@@ -0,0 +1,37 @@
...
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import mate.academy.rickandmorty.model.Character;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import mate.academy.rickandmorty.service.imp.CharacterHttpClient;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class DbFiller {
+    private static final String URL = "https://rickandmortyapi.com/api/character?page=1";
```
vovasalyha, line 16: Why does `DbFiller` should be responsible about knowing the URL of rick and morty API?

src/main/java/mate/academy/rickandmorty/service/DbFiller.java
```diff
@@ -0,0 +1,37 @@
...
+    private final CharacterMapper characterMapper;
+    private final CharacterHttpClient rickAndMortyClient;
+    private final CharacterRepository characterRepository;
+
+    @PostConstruct
+    public void loadDataToDb() {
+        String nextUrl = URL;
+        while (nextUrl != null) {
+            ApiResponseDto apiResponse = rickAndMortyClient.get(nextUrl, ApiResponseDto.class);
```
vovasalyha, line 25: Why does `DbFiller` should be responsible for choosing `ApiResponseDto` as a result type for API?

vovasalyha (changes requested):
Looks good, but see a couple of comments. Also, fix checkstyle errors to fix a build, please.

PR #126, opened 2024-03-04
feat: create Api

dmytro-senko:


src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+@RequiredArgsConstructor
+@Service
+public class CharacterServiceImpl implements CharacterService {
+    private final CharacterRepository characterRepository;
+    private final CharacterMapper characterMapper;
+
+    @Override
+    public CharacterDto getRandomCharacter() {
+        Random random = new Random();
+        Long id = random.nextLong(LoadingExternalDataService.getCountCharacters()) + 1;
```
olekskov, line 25: I'd avoid using static class variable as it may be reason of some unexpected behaviour.
Use repository.count() method instead, it's even simpler

pom.xml
```diff
@@ -66,9 +93,30 @@
...
+							<version>${mapstruct.version}</version>
+						</path>
+					</annotationProcessorPaths>
 				</configuration>
 			</plugin>
 		</plugins>
 	</build>
-
-</project>
+</project>
```
olekskov: empty line in the end of file

olekskov (approved):
Good job, comments below left for consideration

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
dmytro-senko: I reduce the number of queries to the database and thus improve performance.

PR #129, opened 2024-03-12
All tasks were completed

BakhmetIvan:


src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
```diff
@@ -0,0 +1,27 @@
...
+@RequestMapping(value = "/characters")
+public class CharacterController {
+    private final CharacterService characterService;
+
+    @GetMapping("/random")
+    public CharacterDto getRandomCharacter() {
+        return characterService.getRandomCharacter();
+    }
+
+    @GetMapping("by-name")
```
okuzan:
```suggestion
    @GetMapping

```

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+
+@Service
+@RequiredArgsConstructor
+public class CharacterServiceImpl implements CharacterService {
+    private final CharacterMapper characterMapper;
+    private final CharacterDao characterDao;
+
+    @Override
+    public CharacterDto getRandomCharacter() {
+        Random random = new Random();
```
okuzan: don't recreate random needlessly

src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
okuzan: In RESTful API design, it's common to use query parameters to filter collections rather than specifying the filter in the path.

okuzan (changes requested):

Vasul-Pankiv (approved):

Vasul-Pankiv (approved):

PR #135, opened 2024-03-20
implement get random character

hlibmurphy:


kozhukhovsky (changes requested):
We need to fix the build. And also, first, when you launch the application, you need to load all the characters into the database. And when accessing your controller, you should go to the database and pull out data, and not make a request to someone else’s API every time.

src/main/java/mate/academy/rickandmorty/dto/external/ExternalCharacterResponseDto.java
```diff
@@ -0,0 +1,16 @@
...
+import mate.academy.rickandmorty.dto.internal.CharacterResponseDto;
+import org.springframework.stereotype.Component;
+
+@AllArgsConstructor
+@Data
+@Component
+public class ExternalCharacterResponseDto {
+    private InfoResponseDto info;
+    private List<CharacterResponseDto> results;
+
```
okuzan, line 15:
```suggestion
```

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+@AllArgsConstructor
+@Service
+public class CharacterServiceImpl implements CharacterService {
+    private CharacterMapper mapper;
+    private CharacterRepository characterRepository;
+
+    @Override
+    public CharacterDto getRandomCharacterDto() {
+        long size = characterRepository.count();
+        Random random = new Random();
```
okuzan, line 22: better to not recreate random once per method call, but once per class/instance creation

okuzan (approved):
Good job!

PR #137, opened 2024-03-29
Implemented random character generator, and ability to search by name

kacper-kukula:


src/main/java/mate/academy/rickandmorty/dto/external/CharacterResponseDataDto.java
```diff
@@ -0,0 +1,11 @@
+package mate.academy.rickandmorty.dto.external;
+
+import java.util.List;
+import lombok.Data;
+
+@Data
+public class CharacterResponseDataDto {
+
```
nowickimj: 1. Some of your DTO classes are records and some are regular classes - why is that?
2. Unnecessary new line here - please keep the code consistent :)

src/main/java/mate/academy/rickandmorty/service/ApiClient.java
```diff
@@ -0,0 +1,68 @@
...
+            String nextPageUrl = BASE_URL + "/character?page=" + currentPage;
+            CharacterResponseDataDto dataDto = fetchDataFromUrl(nextPageUrl);
+            saveEachCharacterToDb(dataDto.getResults());
+
+            totalPages = dataDto.getInfo().getPages();
+            currentPage++;
+        } while (currentPage <= totalPages);
+    }
+
+    private void saveEachCharacterToDb(List<CharacterDto> results) {
```
nowickimj: `saveAll(List<CharacterDto> characters)`

src/main/java/mate/academy/rickandmorty/service/ApiClient.java
```diff
@@ -0,0 +1,68 @@
...
+public class ApiClient {
+
+    private static final String BASE_URL = "https://rickandmortyapi.com/api";
+
+    private final ObjectMapper objectMapper;
+    private final CharacterMapper characterMapper;
+    private final CharacterRepository characterRepository;
+
+    @PostConstruct
+    public void getAllCharactersToDb() {
```
nowickimj: `fetchAndSaveAllCharacters` -  "toDb" suffix is not necessary

nowickimj (changes requested):

src/main/java/mate/academy/rickandmorty/service/ApiClient.java
```diff
@@ -0,0 +1,68 @@
...
+    private void saveAll(List<CharacterDto> results) {
+
+        List<Character> list = results.stream()
+                .map(characterMapper::toModel)
+                .toList();
+
+        characterRepository.saveAll(list);
+    }
+
+    private CharacterResponseDataDto fetchDataFromUrl(String nextPageUrl) {
```
nowickimj: Few remarks regarding this method.
1. Name is `fetchDataFromUrl` whereas you know what data you're dealing with since you have a fixed return type - so "fetchCharacters" should be enough.
2. You know what type of data are you fetching and what API you want to call - so we don't need a url as an arg here. Please note that the real endpoint url that you're trying to hit is `https://rickandmortyapi.com/api/characters` with `page` passed as a request param (preceded by `?` symbol). So how about a page number as an argument?
3. You're creating a new HttpClient every time you call this method - please make it a dependency

nowickimj (changes requested):
One remark about the method.

src/main/java/mate/academy/rickandmorty/dto/external/CharacterResponseDataDto.java
```diff
@@ -0,0 +1,8 @@
+package mate.academy.rickandmorty.dto.external;
+
+import java.util.List;
+
+public record CharacterResponseDataDto(
```
nowickimj: Didn't catch that before, but you might come up with better name for that class: `GetCharactersResponseDto` - since the operation is GET and you're getting multiple characters, not just one. "Data" word is redundant, it's included in DTO acronym :)

src/main/java/mate/academy/rickandmorty/service/ApiClient.java
```diff
@@ -0,0 +1,67 @@
...
+import mate.academy.rickandmorty.dto.external.CharacterDto;
+import mate.academy.rickandmorty.dto.external.CharacterResponseDataDto;
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import mate.academy.rickandmorty.model.Character;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class ApiClient {
```
nowickimj: This name is too generic. `RickAndMortyApi` would be more explicit.

src/main/java/mate/academy/rickandmorty/service/ApiClient.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private static final HttpClient HTTP_CLIENT = HttpClient.newHttpClient();
+    private static final String BASE_URL = "https://rickandmortyapi.com/api/character?page=";
```
nowickimj: As I've mentioned, url to characters endpoint is `https://rickandmortyapi.com/api/character`.

HOWEVER, since you're creating the general RickAndMorty API client - the real base url should be `https://rickandmortyapi.com/api` and `/character` path only to be added on endpoint call method.

`page` is a request param so it should be added on `HttpRequest` building stage, not hardcoded in url.

src/main/java/mate/academy/rickandmorty/service/ApiClient.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private final ObjectMapper objectMapper;
+    private final CharacterMapper characterMapper;
+    private final CharacterRepository characterRepository;
+
+    @PostConstruct
+    public void getAllCharacters() {
+        int currentPage = 1;
```
nowickimj: You can make this solution a bit clearer:
1. Call endpoint url (`/api/characters`) - check for `next` in `results`.
2. Until `next` exists (not a null), keep fetching results from `next`, updating its value every time.

nowickimj (changes requested):
Your approach on designing API client a bit different to what I proposed. Also I found few more things to fix which I've overlooked before - please take a look.

src/main/java/mate/academy/rickandmorty/service/ApiClient.java
nowickimj: PS:
hint for designing API call method: `https://rickandmortyapi.com/api/character` is equivalent to `https://rickandmortyapi.com/api/character?page=1`, so you can either make page parameter optional (add only when it's provided as arg) or set `1` by default. In my opinion, option #1 will be better.

kacper-kukula: I'm not sure if I understood that correctly, but if the page parameter is optional, then how do we get characters from all the other pages? Or you just meant that for the first iteration leave it without paramater and then just begin with page=2 and so on?

kacper-kukula:
At the end I figured that the page parameter is redundant, because we get the full link to the next page from the response anyway , so I used that and shortened the code by few lines.

src/main/java/mate/academy/rickandmorty/service/ApiClient.java
nowickimj: Yes, I was referring to first call only. 
BUT to my surprise (haven't used Java's new HttpClient before, just libraries), THERE IS NO PARAMS SETTERS AVAILABLE which seems ridiculous to me.

So I guess, in case of url building, we would need to deal with hardcoding the `?page=` in url, however I'd still stick with separating `BASE_URL` from `getCharactersPath` (`/characters`) and page param value (`?page=`) - join this values in API call method.

EDIT: this comment is irrelevant since you started using `next` value :)

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApi.java
```diff
@@ -0,0 +1,65 @@
...
+    private static final HttpClient HTTP_CLIENT = HttpClient.newHttpClient();
+    private static final String BASE_URL = "https://rickandmortyapi.com/api";
+
+    private final ObjectMapper objectMapper;
+    private final CharacterMapper characterMapper;
+    private final CharacterRepository characterRepository;
+
+    @PostConstruct
+    public void getAllCharacters() {
+        String urlToNextPage = BASE_URL + "/character";
```
nowickimj: Just `requestUrl` or `currentRequestUrl` - it's not the next page at the beginning.

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApi.java
```diff
@@ -0,0 +1,65 @@
...
+
+        do {
```
nowickimj: since you know the initial value, `while` would be make more sense.

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApi.java
```diff
@@ -0,0 +1,65 @@
...
+            GetCharactersResponseDto dataDto = fetchCharacters(urlToNextPage);
```
nowickimj: `response`, please avoid using such generic names.

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApi.java
```diff
@@ -0,0 +1,65 @@
...
+            saveAll(dataDto.results());
+
+            urlToNextPage = dataDto.info().next();
+        } while (urlToNextPage != null);
+    }
+
+    private void saveAll(List<CharacterDto> results) {
+
```
nowickimj: Unnecessary new line

src/main/java/mate/academy/rickandmorty/service/RickAndMortyApi.java
```diff
@@ -0,0 +1,65 @@
...
+        List<Character> list = results.stream()
+                .map(characterMapper::toModel)
+                .toList();
+
+        characterRepository.saveAll(list);
+    }
+
+    private GetCharactersResponseDto fetchCharacters(String urlToNextPage) {
```
nowickimj: "next" doesn't make sense in a context of this method - the API client doesn't know if it's first, second or next, he just accepts the `url` and performs a call.

nowickimj (changes requested):

src/main/java/mate/academy/rickandmorty/service/CharacterService.java
```diff
@@ -0,0 +1,36 @@
...
+public class CharacterService {
+
+    private static final Random RANDOM_GENERATOR = new Random();
+
+    private final CharacterRepository characterRepository;
+    private final CharacterMapper characterMapper;
+
+    public CartoonCharacterDto getRandomCharacter() {
+        long entriesCount = characterRepository.count() + 1L;
+        long randomId = RANDOM_GENERATOR.nextLong(entriesCount);
```
okuzan, line 23:
```suggestion
        long entriesCount = characterRepository.count();
        long randomId = RANDOM_GENERATOR.nextLong(entriesCount + 1L);
```
to be a bit more precise

okuzan (approved):
LGTM

PR #142, opened 2024-04-06
hw solution

SkillfulDev:


src/main/java/mate/academy/rickandmorty/dto/external/ExternalCharacterResponseDto.java
```diff
@@ -0,0 +1,14 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import lombok.Data;
+import mate.academy.rickandmorty.entity.Character;
+
+@Data
+public class ExternalCharacterResponseDto {
+    @JsonProperty("results")
+    private List<Character> characters;
+    @JsonProperty("info")
```
andrii-hoienko: What is the reason to add JsonProperty annotations here?

src/main/java/mate/academy/rickandmorty/dto/external/PageInfo.java
```diff
@@ -0,0 +1,16 @@
...
+@Data
+public class PageInfo {
+    @JsonProperty("next")
+    private String next;
+    @JsonProperty("prev")
+    private String prev;
+    @JsonProperty("pages")
+    private int pages;
+    @JsonProperty("count")
+    private int count;
```
andrii-hoienko: What is the reason to add JsonProperty annotations here?

src/main/java/mate/academy/rickandmorty/entity/Character.java
```diff
@@ -0,0 +1,31 @@
...
+import jakarta.persistence.Table;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
+@Getter
+@Setter
+@RequiredArgsConstructor
+@ToString
```
andrii-hoienko:
```suggestion
@Data
```

src/main/java/mate/academy/rickandmorty/mapper/impl/CharacterMapperImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+import mate.academy.rickandmorty.dto.internal.CharacterResponseDto;
+import mate.academy.rickandmorty.entity.Character;
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import org.springframework.stereotype.Component;
+
+@Component
+public class CharacterMapperImpl implements CharacterMapper {
+    public CharacterMapperImpl() {
+    }
```
andrii-hoienko:
```suggestion
```

src/main/java/mate/academy/rickandmorty/mapper/impl/CharacterMapperImpl.java
```diff
@@ -0,0 +1,41 @@
...
+                characterResponseDto.setStatus(character.getStatus());
+            }
+
+            if (character.getGender() != null) {
+                characterResponseDto.setGender(character.getGender());
+            }
+
+            return characterResponseDto;
+        }
+    }
```
andrii-hoienko: I dont understand why such logic is required. If the field is null we wont set it, but all not initialized fields is null by default anyway

src/main/java/mate/academy/rickandmorty/service/impl/ExternalClient.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.List;
+import lombok.RequiredArgsConstructor;
+import mate.academy.rickandmorty.dto.external.ExternalCharacterResponseDto;
+import mate.academy.rickandmorty.entity.Character;
+import org.springframework.stereotype.Component;
+
+@RequiredArgsConstructor
+@Component
+public class ExternalClient {
+    private static final String URL = "https://rickandmortyapi.com/api/character";
```
andrii-hoienko: I believe this value should be taken from properties file, because URL could change and you can make request on different env of external API, so better to make it configurable

src/main/java/mate/academy/rickandmorty/service/impl/ExternalClient.java
```diff
@@ -0,0 +1,50 @@
...
+    private final ObjectMapper objectMapper;
+
+    public List<Character> getAllCharacters() {
+        List<Character> allCharacters = new ArrayList<>();
+        HttpClient httpClient = HttpClient.newHttpClient();
+        String nextUrl = URL;
+
+        while (nextUrl != null && !nextUrl.isEmpty()) {
```
andrii-hoienko:
```suggestion
        while (nextUrl != null && !nextUrl.isBlank()) {
```

src/main/java/mate/academy/rickandmorty/service/impl/ExternalClient.java
```diff
@@ -0,0 +1,50 @@
...
+
+@RequiredArgsConstructor
+@Component
+public class ExternalClient {
+    private static final String URL = "https://rickandmortyapi.com/api/character";
+    private final ObjectMapper objectMapper;
+
+    public List<Character> getAllCharacters() {
+        List<Character> allCharacters = new ArrayList<>();
+        HttpClient httpClient = HttpClient.newHttpClient();
```
andrii-hoienko: Are you sure we need to create this object each time method is called?

src/main/java/mate/academy/rickandmorty/service/impl/ExternalClient.java
```diff
@@ -0,0 +1,50 @@
...
+                );
+                ExternalCharacterResponseDto pageData = objectMapper.readValue(
+                        response.body(),
+                        ExternalCharacterResponseDto.class
+                );
+
+                allCharacters.addAll(pageData.getCharacters());
+                nextUrl = pageData.getInfo().getNext();
+            } catch (IOException | InterruptedException e) {
+                throw new RuntimeException("Can't get response from source", e);
```
andrii-hoienko: Here 2 exceptions could occur, first one is related to mapping and second one is related to invalid request so they are different wery much, I suggest to create 2 separate catch blocks with different error messages

andrii-hoienko: Create custom exception for each case I described

src/main/java/mate/academy/rickandmorty/service/impl/RickAndMortyServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+import mate.academy.rickandmorty.entity.Character;
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import mate.academy.rickandmorty.repository.RickAndMortyRepository;
+import mate.academy.rickandmorty.service.RickAndMortyService;
+import org.springframework.stereotype.Service;
+
+@RequiredArgsConstructor
+@Service
+public class RickAndMortyServiceImpl implements RickAndMortyService {
+    private static final Random random = new Random();
```
andrii-hoienko: make it just class level variable or consider proper naming

src/main/java/mate/academy/rickandmorty/service/impl/RickAndMortyServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+    private final RickAndMortyRepository rickAndMortyRepository;
+    private final CharacterMapper characterMapper;
+
+    @Override
+    public CharacterResponseDto getRandomCharacter() {
+        long maxGeneratedNumber = rickAndMortyRepository.count() + 1;
```
andrii-hoienko: magic number

src/main/java/mate/academy/rickandmorty/service/impl/RickAndMortyServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+        return characterMapper.toDto(rickAndMortyRepository.findById(
+                        random.nextLong(1, maxGeneratedNumber)).orElseThrow(() ->
+                        new RuntimeException("Can't find Character in DB")));
```
andrii-hoienko: Bad exception, you can use more suitable or create your own

src/main/java/mate/academy/rickandmorty/service/impl/RickAndMortyServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+public class RickAndMortyServiceImpl implements RickAndMortyService {
+    private static final Random random = new Random();
+    private final RickAndMortyRepository rickAndMortyRepository;
+    private final CharacterMapper characterMapper;
+
+    @Override
+    public CharacterResponseDto getRandomCharacter() {
+        long maxGeneratedNumber = rickAndMortyRepository.count() + 1;
+        return characterMapper.toDto(rickAndMortyRepository.findById(
+                        random.nextLong(1, maxGeneratedNumber)).orElseThrow(() ->
```
andrii-hoienko: I suggest create method that will return random character id

src/main/resources/application.properties
```diff
@@ -1 +1,10 @@
+spring.datasource.url= jdbc:mysql://localhost:3306/rick_and_morty
+spring.datasource.username=${DB_USERNAME}
+spring.datasource.password=${DB_PASSWORD}
```
andrii-hoienko, line 3: Good job!

src/main/java/mate/academy/rickandmorty/service/impl/ExternalClient.java
```diff
@@ -0,0 +1,50 @@
...
+import mate.academy.rickandmorty.entity.Character;
+import org.springframework.stereotype.Component;
+
+@RequiredArgsConstructor
+@Component
+public class ExternalClient {
+    private static final String URL = "https://rickandmortyapi.com/api/character";
+    private final ObjectMapper objectMapper;
+
+    public List<Character> getAllCharacters() {
```
andrii-hoienko, line 27: All data from the public API should be fetched once, and only once, when the Application context is created

andrii-hoienko (changes requested):

src/main/java/mate/academy/rickandmorty/dto/external/ExternalCharacterResponseDto.java
SkillfulDev: I removed these annotation and change "characters" onto results to have matches with JSON file that we receive

src/main/java/mate/academy/rickandmorty/dto/external/PageInfo.java
SkillfulDev: I thought we have to use JsonProperty by default even when name of data from JSON equal to our java naming

src/main/java/mate/academy/rickandmorty/entity/Character.java
SkillfulDev: I changed it , but I have this warning 
image

src/main/java/mate/academy/rickandmorty/mapper/impl/CharacterMapperImpl.java
SkillfulDev: fixed problems with mupstruct

src/main/java/mate/academy/rickandmorty/mapper/impl/CharacterMapperImpl.java
SkillfulDev: fixed problems with mupstruct

src/main/java/mate/academy/rickandmorty/service/impl/ExternalClient.java
SkillfulDev: Changed. I call Thread.currentThread().interrupt () to keep the break status because InterruptedException clears the break status.

src/main/java/mate/academy/rickandmorty/service/impl/ExternalClient.java
SkillfulDev: This rules is executed in InitializationData

src/main/java/mate/academy/rickandmorty/entity/Character.java
karma-o: Yes, I actually agree with the warning, better to use the one that you had before, but we also should not forget about equals & hashcode

src/main/java/mate/academy/rickandmorty/controller/RickAndMortyController.java
```diff
@@ -0,0 +1,33 @@
...
+    @GetMapping
+    @Operation(summary = "Get a random character", description = "Get a random character")
+    public CharacterResponseDto getRandomCharacter() {
+        return rickAndMortyService.getRandomCharacter();
+    }
+
+    @GetMapping("/params")
+    @Operation(summary = "Get all characters contain putting string into parameter name",
+            description = "Get all characters contain putting string into parameter name")
+    public List<CharacterResponseDto> getByName(@RequestParam(name = "name") String name) {
```
karma-o, line 30: we could also add a bit more of a `name` param documentation to swagger

karma-o (approved):
Looks good, left a comment on the `@Data` warning and another minor one

PR #144, opened 2024-04-13
Created api for getting "Rick and Morty" characters info

k0sm0naft:


src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
```diff
@@ -0,0 +1,35 @@
...
+    @Operation(summary = "Return character info",
+            description = "Return single random character info")
+    public CharacterDto getRandomCharacter() {
+        return characterService.getRandomCharacter();
+    }
+
+    @GetMapping("/{word}")
+    @Operation(summary = "Return list of characters",
+            description = "Return list of characters info that contains word in their names")
+    public List<CharacterDto> searchByName(@PathVariable String word) {
```
karma-o: documentation for `word` param would be nice to have in swagger, and maybe rename to `name`

src/main/java/mate/academy/rickandmorty/dto/external/InfoDto.java
```diff
@@ -0,0 +1,9 @@
+package mate.academy.rickandmorty.dto.external;
+
+public record InfoDto(
```
karma-o: this is more of a page info, just info is too abstract

src/main/java/mate/academy/rickandmorty/model/Character.java
```diff
@@ -0,0 +1,32 @@
...
+import jakarta.persistence.Table;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@Entity
+@Table(name = "characters")
```
karma-o: what about the equals and hashcode?

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.stereotype.Service;
+
+@RequiredArgsConstructor
+@Service
+public class CharacterServiceImpl implements CharacterService {
+    private final CharacterDao characterDao;
+    private final CharacterMapper characterMapper;
+    private final RickAndMortyCharactersClient charactersClient;
+    private final Random random = new Random();
+    private int countOfCharacters;
```
karma-o: I wouldn't do that, what if we call `getRandomCharacter()` before `saveAllExternalCharacters()` is called, and in general seems like a bad idea, to store something like that as a service field

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+        List<Character> characters = charactersClient.getExternalCharacters().stream()
+                                               .map(characterMapper::toModel)
+                                               .toList();
+        countOfCharacters = characters.size();
+        characterDao.saveAll(characters);
+    }
+
+    @Override
+    public CharacterDto getRandomCharacter() {
+        int randomId = random.nextInt(countOfCharacters);
```
karma-o: just do another call to count all rows in the `characters` table, and it would be fine

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+        return characterDao.findById(randomId)
+                           .map(characterMapper::toDto)
+                           .orElseThrow(() -> new NoSuchElementException(
+                                   "Can't get random character by id " + randomId));
+    }
+
+    @Override
+    public List<CharacterDto> findByName(String name) {
+        return characterDao.findAllByNameContains(name).stream()
```
karma-o, line 52: we can try to make it work ignorecase, if you'd like

src/main/java/mate/academy/rickandmorty/service/RickAndMortyCharactersClient.java
```diff
@@ -0,0 +1,53 @@
...
+    private String baseUrl;
+
+    public List<ExternalCharacterDto> getExternalCharacters() {
+        List<ExternalCharacterDto> externalCharacters = new ArrayList<>();
+
+        while (baseUrl != null) {
+            ExternalDto externalDto = getExternalDto(baseUrl);
+            externalCharacters.addAll(externalDto.results());
+            baseUrl = externalDto.info().next();
+        }
```
karma-o: you can make a validation of that `baseUrl` in `@PostConstruct` to ensure that it is valid on the start of the application, instead of when you have already called the method

src/main/java/mate/academy/rickandmorty/service/RickAndMortyCharactersClient.java
```diff
@@ -0,0 +1,53 @@
...
+        HttpClient httpClient = HttpClient.newHttpClient();
+        HttpRequest httpRequest = HttpRequest.newBuilder()
+                                             .GET()
+                                             .uri(URI.create(url))
+                                             .build();
+        try {
+            HttpResponse<String> response =
+                    httpClient.send(httpRequest, HttpResponse.BodyHandlers.ofString());
+            return objectMapper.readValue(response.body(), ExternalDto.class);
+        } catch (IOException | InterruptedException e) {
```
karma-o: I'd rather split that catch block and handling into 2

src/main/java/mate/academy/rickandmorty/service/RickAndMortyCharactersClient.java
```diff
@@ -0,0 +1,53 @@
...
+        while (baseUrl != null) {
+            ExternalDto externalDto = getExternalDto(baseUrl);
+            externalCharacters.addAll(externalDto.results());
+            baseUrl = externalDto.info().next();
+        }
+
+        return externalCharacters;
+    }
+
+    private ExternalDto getExternalDto(String url) {
```
karma-o: we do not use that method with any other URL so far, so we can use `baseUrl` directly without passing it as a method param

src/main/resources/application.properties
```diff
@@ -1 +1,10 @@
+spring.datasource.url=jdbc:mysql://localhost:3306/rick_and_morty?serverTimezone=UTC
+spring.datasource.driver-class-name=com.mysql.cj.jdbc.Driver
+spring.datasource.username=<USER_NAME>
+spring.datasource.password=<PASSWORD>
 
+spring.jpa.hibernate.ddl-auto=create-drop
+spring.jpa.show-sql=true
+spring.jpa.open-in-view=false
+
+api.characters.url=https://rickandmortyapi.com/api/character
```
karma-o: I'd suggest setting only the domain here, and adding the actual endpoint path in the `RickAndMortyCharactersClient` to have more flexibility in case we will want to add another fetсh method later on

src/main/java/mate/academy/rickandmorty/service/RickAndMortyCharactersClient.java
```diff
@@ -0,0 +1,53 @@
+package mate.academy.rickandmorty.service;
```
karma-o:
```suggestion
package mate.academy.rickandmorty.client;
```

karma-o (changes requested):
Looks good! left several comments

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
k0sm0naft: `findAllByNameContains` already work with ignore case.

pom.xml
```diff
@@ -41,6 +43,34 @@
...
+		<dependency>
+			<groupId>org.mapstruct</groupId>
+			<artifactId>mapstruct</artifactId>
+			<version>${org.mapstruct.version}</version>
+		</dependency>
+
+		<dependency>
+			<groupId>org.springdoc</groupId>
+			<artifactId>springdoc-openapi-starter-webmvc-ui</artifactId>
+			<version>2.5.0</version>
```
andrii-hoienko: Please use variables for versions as you did in ${org.mapstruct.version}

src/main/java/mate/academy/rickandmorty/client/RickAndMortyCharactersClient.java
```diff
@@ -0,0 +1,57 @@
...
+            ExternalDto externalDto = getExternalDto();
+            externalCharacters.addAll(externalDto.results());
+            fullUrl = externalDto.info().next();
+        }
+
+        return externalCharacters;
+    }
+
+    private ExternalDto getExternalDto() {
+        HttpClient httpClient = HttpClient.newHttpClient();
```
andrii-hoienko: It would be great if you will create bean of HttpClient in some config file, and than Autowire it in all Clients

src/main/java/mate/academy/rickandmorty/client/RickAndMortyCharactersClient.java
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@RequiredArgsConstructor
+@Component
+public class RickAndMortyCharactersClient {
+    private final ObjectMapper objectMapper;
+    @Value("${api.rick.and.morty.url}")
+    private String baseUrl;
+    private String fullUrl;
```
andrii-hoienko: Please make it local variable, it is possible to pass fullUrl as a method argument

src/main/java/mate/academy/rickandmorty/client/RickAndMortyCharactersClient.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public List<ExternalCharacterDto> getExternalCharacters() {
+        List<ExternalCharacterDto> externalCharacters = new ArrayList<>();
+        fullUrl = baseUrl + "/api/character";
```
andrii-hoienko: Please use constants for string literals, but I believe that '/api/character' could be added to baseUrl

src/test/resources/application.properties
```diff
@@ -3,3 +3,5 @@ spring.datasource.driverClassName=org.h2.Driver
 spring.datasource.username=sa
 spring.datasource.password=password
 spring.jpa.database-platform=org.hibernate.dialect.H2Dialect
+
+api.rick.and.morty.url=https://rickandmortyapi.com
```
andrii-hoienko:
```suggestion
rickandmorty.url=https://rickandmortyapi.com
```

src/main/java/mate/academy/rickandmorty/client/RickAndMortyCharactersClient.java
```diff
@@ -0,0 +1,57 @@
...
+import lombok.RequiredArgsConstructor;
+import mate.academy.rickandmorty.dto.external.ExternalCharacterDto;
+import mate.academy.rickandmorty.dto.external.ExternalDto;
+import mate.academy.rickandmorty.exception.InvalidApiResponseException;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@RequiredArgsConstructor
+@Component
+public class RickAndMortyCharactersClient {
```
andrii-hoienko, line 16: OPTIONAL: Actually client should only make requests, logic related to URl calculation should be in separate class for example some RickAndMortyCharactersService

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+
+@RequiredArgsConstructor
+@Service
+public class CharacterServiceImpl implements CharacterService {
+    private final CharacterDao characterDao;
+    private final CharacterMapper characterMapper;
+    private final RickAndMortyCharactersClient charactersClient;
+    private final Random random = new Random();
+
+    @PostConstruct
```
andrii-hoienko: OPTIONAL: PostConstruct is about setting up and configuring beans, not for business logic execution, there are other ways to execute code on startup like EventListeners or CommandLineRunner interface, I would prefer to use them

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+import mate.academy.rickandmorty.repository.CharacterDao;
+import org.springframework.stereotype.Service;
+
+@RequiredArgsConstructor
+@Service
+public class CharacterServiceImpl implements CharacterService {
+    private final CharacterDao characterDao;
+    private final CharacterMapper characterMapper;
+    private final RickAndMortyCharactersClient charactersClient;
+    private final Random random = new Random();
```
andrii-hoienko: Better to create bean of Random and autowire it

andrii-hoienko (changes requested):

src/main/java/mate/academy/rickandmorty/client/RickAndMortyCharactersClient.java
k0sm0naft: @karma-o suggested setting only the domain in `baseUrl`, and adding the actual endpoint path in the RickAndMortyCharactersClient to have more flexibility in case we will want to add another fetch method later on

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,56 @@
...
+import mate.academy.rickandmorty.repository.CharacterDao;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.event.ContextRefreshedEvent;
+import org.springframework.context.event.EventListener;
+import org.springframework.stereotype.Service;
+
+@RequiredArgsConstructor
+@Service
+public class CharacterServiceImpl implements CharacterService {
+    public static final String ENDPOINT = "/api/character";
```
karma-o:
```suggestion
    public static final String CHARACTERS_ENDPOINT_PATH = "/api/character";
```

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,56 @@
...
+    private final CharacterDao characterDao;
+    private final CharacterMapper characterMapper;
+    private final RickAndMortyCharactersClient charactersClient;
+    private final Random random;
+    @Value("${rickandmorty.url}")
+    private String baseUrl;
+
+    @EventListener(ContextRefreshedEvent.class)
```
karma-o: https://stackoverflow.com/a/67227144/13842504
```suggestion
    @EventListener(ApplicationReadyEvent.class)
```

karma-o (approved):
Good job! Also left a couple suggestions

PR #147, opened 2024-04-19
external api rickandmortyapi.com/api/character added

maiiat:
1) External rickandmortyapi.com/api/character API Service added
2) getRandomCharacterter() and getCharacterByName() methods implemented


koval-kn (changes requested):
checks failed, please fix them first
image

olekskov (approved):
Nice job!
Probably may use saveAll operation during characters fetch to decrease number of calls for DB

PR #149, opened 2024-04-26
Hw jv rick and morty

AdamMudrak:


src/main/java/mate/academy/rickandmorty/Application.java
```diff
@@ -1,12 +1,68 @@
...
+     */
+    private void openApiInBrowser() {
+        System.setProperty(HEADLESS_PROPERTY, HEADLESS_PROPERTY_VALUE);
+        Desktop desktop = Desktop.getDesktop();
+        try {
+            desktop.browse(new URI(URL_OF_API));
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
```
koval-kn: just use okhttp for creating requests

src/main/java/mate/academy/rickandmorty/dto/internal/RemoteToLocalDto.java
```diff
@@ -0,0 +1,11 @@
+package mate.academy.rickandmorty.dto.internal;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public record RemoteToLocalDto(
```
koval-kn, line 5: why do you have 2 DTOs for internal and extra 2 for external? try to use only 1 per each type

src/main/java/mate/academy/rickandmorty/service/CharacterRemoteClient.java
```diff
@@ -0,0 +1,67 @@
...
+import mate.academy.rickandmorty.dto.external.CharacterResponseDataDto;
+import mate.academy.rickandmorty.dto.internal.RemoteToLocalDto;
+import org.springframework.stereotype.Component;
+
+@RequiredArgsConstructor
+@Component
+public class CharacterRemoteClient {
+    private static final String BASE_URL = "https://rickandmortyapi.com/api/character?page=%d";
+    private static final Integer DEFAULT_PAGE = 1;
+    private static final HttpClient HTTP_CLIENT = HttpClient.newHttpClient();
```
koval-kn, line 24: use okhttp, its easier

koval-kn (changes requested):

src/main/java/mate/academy/rickandmorty/dto/internal/RemoteToLocalDto.java
AdamMudrak: I have a few reasons for having 2 DTOS of each type.

I need to map two fields containing subfields from server - those are **info** and **results**.

**CharacterInfoDto info** is vital for me as its subfield, next, is needed to be able to iterate over each page existing now and in future of this remote API.

**RemoteToLocalDto[] results** is nothing to explain, I just need it to map each character locally.

Plus I want to divide meta data from client data more obviously, just like Chupika did in the video.

But what about **RestClientDto**? Thing is, RemoteToLocalDto contains, logically, only one ID field which is received from third party API, so I can't use it to send a response to my API client as it will show only one ID field, which will not necessarily be always the same for both externalId and id in my local database.

I also would like to point out that I checked other students' approved solutions, and they definitely have more than 2 DTOs, or even more models

src/main/java/mate/academy/rickandmorty/Application.java
AdamMudrak: I'll just delete it then. Honestly, I'm not ready to spend time on learning a new library for optional tasks.

src/main/java/mate/academy/rickandmorty/service/CharacterRemoteClient.java
AdamMudrak: I don't think it is since it was not shown in the course at all. I don't know its methods and logic of work. If possible, I'd like to stick to Http, just like most of the others.

sarakhmen (approved):
LGTM!

PR #151, opened 2024-05-06
init commit

TonyH277:


src/main/java/mate/academy/rickandmorty/config/DataFetcher.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private List<Character> deserializeCharacters(JsonNode rootNode)
+            throws JsonProcessingException {
+        JsonNode resultNode = rootNode.get(ROOT_RESULTS);
+        return objectMapper.treeToValue(resultNode,
+                objectMapper.getTypeFactory().constructCollectionType(List.class, Character.class));
+
+    }
+}
+
```
andrii-hoienko:
```suggestion
```

src/main/java/mate/academy/rickandmorty/config/DataFetcher.java
```diff
@@ -0,0 +1,60 @@
...
+    private final ObjectMapper objectMapper;
+    private final CharacterRepository characterRepository;
+
+    @PostConstruct
+    public void fetchData() throws JsonProcessingException {
+        String url = CHARACTER_URL;
+
+        while (url != null) {
+            String jsonResponse = restTemplate.getForObject(url, String.class);
+            JsonNode rootNode = objectMapper.readTree(jsonResponse);
```
andrii-hoienko: What is the reason to readTree? Is it possible to map to an object?

src/main/java/mate/academy/rickandmorty/config/DataFetcher.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final String ROOT_INFO = "info";
+    private static final String ROOT_RESULTS = "results";
+    private static final String ROOT_NEXT = "next";
+    private static final String NULL_VALUE = "null";
+    private final RestTemplate restTemplate;
+    private final ObjectMapper objectMapper;
+    private final CharacterRepository characterRepository;
+
+    @PostConstruct
+    public void fetchData() throws JsonProcessingException {
```
andrii-hoienko: Bad method name, I believe fetch data method should return something

andrii-hoienko: Please catch JsonProcessingException and rethrow it as runtime

src/main/java/mate/academy/rickandmorty/config/DataFetcher.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final String CHARACTER_URL = "https://rickandmortyapi.com/api/character";
+    private static final String ROOT_INFO = "info";
+    private static final String ROOT_RESULTS = "results";
+    private static final String ROOT_NEXT = "next";
+    private static final String NULL_VALUE = "null";
+    private final RestTemplate restTemplate;
+    private final ObjectMapper objectMapper;
+    private final CharacterRepository characterRepository;
+
+    @PostConstruct
```
andrii-hoienko: PostConstruct often used for bean configuration, the code that you wrote in not related to bean configuration, so I suggest to use different approach for this, you can use CommandLineRunner interface(or bean) or use event listeners to run code on startup

src/main/java/mate/academy/rickandmorty/config/RestTemplateConfig.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Bean
+    public CloseableHttpClient httpClient() {
+        PoolingHttpClientConnectionManager connManager = new PoolingHttpClientConnectionManager();
+        connManager.setMaxTotal(100);
+        connManager.setDefaultMaxPerRoute(20);
+        return HttpClients.custom()
+                .setConnectionManager(connManager)
+                .build();
+    }
```
andrii-hoienko, line 30: This beans should be created by spring, no need to create them manually, if you want to configure them somehow you can you application properties

src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
```diff
@@ -0,0 +1,37 @@
...
+import mate.academy.rickandmorty.service.CharacterService;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+@Tag(name = "Rick and Morty character management",
+        description = "Endpoints for managing characters")
+@RestController
+@RequestMapping("characters")
```
andrii-hoienko:
```suggestion
@RequestMapping("/characters")
```

src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
```diff
@@ -0,0 +1,37 @@
...
+@RequiredArgsConstructor
+public class CharacterController {
+    private final CharacterService characterService;
+
+    @Operation(summary = "Get one character",
+            description = "Get random character from Rick and Morty universe")
+    @GetMapping
+    public Character generate() {
```
andrii-hoienko: bad method name

src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Operation(summary = "Get list of characters by name contains 'row'",
+            description = "Get list of character from Rick and Morty universe "
+                    + "where character name like '%:row%'")
+    @GetMapping("/byNameContains")
+    public List<Character> findCharacterNameContains(@RequestParam String row) {
+        return characterService.findCharacterNameContains(row);
+    }
+
```
andrii-hoienko:
```suggestion
```

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+@RequiredArgsConstructor
+public class CharacterServiceImpl implements CharacterService {
+    private static final int CHARACTERS_AMOUNT = 826;
+    private final CharacterRepository characterRepository;
+
+    @Override
+    public Character generateRandomCharacter() {
+        return Optional.of(characterRepository.findById(new Random().nextLong(CHARACTERS_AMOUNT)))
+                .get()
+                .get();
```
andrii-hoienko: What if optional will be empty?

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+
+@Service
+@RequiredArgsConstructor
+public class CharacterServiceImpl implements CharacterService {
+    private static final int CHARACTERS_AMOUNT = 826;
+    private final CharacterRepository characterRepository;
+
+    @Override
+    public Character generateRandomCharacter() {
+        return Optional.of(characterRepository.findById(new Random().nextLong(CHARACTERS_AMOUNT)))
```
andrii-hoienko: Do you really want to create Random object each time this method is called

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+import lombok.RequiredArgsConstructor;
+import mate.academy.rickandmorty.model.Character;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import mate.academy.rickandmorty.service.CharacterService;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class CharacterServiceImpl implements CharacterService {
+    private static final int CHARACTERS_AMOUNT = 826;
```
andrii-hoienko: What if some characters will be added or removed?

src/main/java/mate/academy/rickandmorty/utils/CharacterDeserializer.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.fasterxml.jackson.core.JacksonException;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonNode;
+import java.io.IOException;
+import mate.academy.rickandmorty.model.Character;
+
+public class CharacterDeserializer extends JsonDeserializer<Character> {
```
andrii-hoienko: I think it is redundant, object mapper could deserialize your objects without additional configuration

src/main/resources/application.properties
```diff
@@ -1 +1,8 @@
-
+spring.application.name=rickandmorty
+spring.datasource.url=jdbc:mysql://localhost:3306/rickandmorty
+spring.datasource.driver-class-name=com.mysql.cj.jdbc.Driver
+spring.datasource.username=root
+spring.datasource.password=root
```
andrii-hoienko: Don`t push createntials

src/main/java/mate/academy/rickandmorty/config/DataFetcher.java
```diff
@@ -0,0 +1,60 @@
...
+import mate.academy.rickandmorty.model.Character;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import org.springframework.stereotype.Component;
+import org.springframework.web.client.RestTemplate;
+
+@Component
+@RequiredArgsConstructor
+public class DataFetcher {
+
+    private static final String CHARACTER_URL = "https://rickandmortyapi.com/api/character";
```
andrii-hoienko: I would prefer to make this url configurable via propertioes

andrii-hoienko (changes requested):

src/main/java/mate/academy/rickandmorty/config/RestTemplateConfig.java
TonyH277: In my case I add those, because spring said
"Parameter 0 of constructor in mate.academy.rickandmorty.config.DataFetcher required a bean of type 'org.springframework.web.client.RestTemplate' that could not be found.


Action:

Consider defining a bean of type 'org.springframework.web.client.RestTemplate' in your configuration.
"

src/main/java/mate/academy/rickandmorty/utils/CharacterDeserializer.java
TonyH277: It does, but it maps id from Json to Entity.id. I thought JsonProperty annotation will solve this problem, but had mapping exception with msg like (I want to set json.id to entity.id, but your annotation said that i should set json.id to other field)

src/main/java/mate/academy/rickandmorty/config/DataFetcher.java
TonyH277: I choose this aproach because its clear for me to create object from which then I'll fetch info by key, and then manipulate with it. Then I get by key="results" and map to Character.class

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    @Transactional
+    public Character generateRandomCharacter() {
+        long charactersAmount = characterRepository.count();
+        long id = random.nextLong(charactersAmount);
+        return Optional.of(characterRepository.findById(id))
+                .get()
+                .orElseThrow(() -> new IllegalStateException("Character not found with id: " + id));
+    }
```
andrii-hoienko: as far as I remember characterRepository.findById(id) returns optional. So you make Optional.of(Optional returned from repository).get (//Unboxing optional that you create) .orElseThrow (//Working with optional returned from repo)

src/main/java/mate/academy/rickandmorty/utils/CharacterDeserializer.java
andrii-hoienko: That is why you need to create an externalDto to map response from API and then remap it to entity and save it, such approach will be more flexible because we have a separate class to represent characters in external API. In general better to avoid custom deserialization logic, because it is often confusing. Flow when we map external response as is and then work with it as we want is preferable, I could describe in more details on QnA

src/main/java/mate/academy/rickandmorty/utils/DataFetcher.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public void populateCharactersDB() {
+        String url = characterUrl;
+
+        while (url != null) {
+            String jsonResponse = restTemplate.getForObject(url, String.class);
+            try {
+                JsonNode rootNode = objectMapper.readTree(jsonResponse);
+                List<Character> pageCharacters = deserializeCharacters(rootNode);
+                characterRepository.saveAll(pageCharacters);
```
andrii-hoienko: I suggest to fetch all characters first and than save them all to not make request to db several times

src/main/java/mate/academy/rickandmorty/utils/DataFetcher.java
```diff
@@ -0,0 +1,61 @@
...
+import lombok.RequiredArgsConstructor;
+import mate.academy.rickandmorty.model.Character;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+import org.springframework.web.client.RestTemplate;
+
+@Component
+@RequiredArgsConstructor
+public class DataFetcher {
```
andrii-hoienko:
```suggestion
public class DataFetchService{
```

src/main/java/mate/academy/rickandmorty/utils/DataFetcher.java
```diff
@@ -0,0 +1,61 @@
...
+    private static final String ROOT_RESULTS = "results";
+    private static final String ROOT_NEXT = "next";
+    private static final String NULL_VALUE = "null";
+    private final RestTemplate restTemplate;
+    private final ObjectMapper objectMapper;
+    private final CharacterRepository characterRepository;
+    @Value("${characters.url}")
+    private String characterUrl;
+
+    public void populateCharactersDB() {
```
andrii-hoienko:
```suggestion
    public List<Characters> fetch() {
```

src/main/java/mate/academy/rickandmorty/config/CustomApplicationRunner.java
```diff
@@ -0,0 +1,18 @@
...
+@Component
+@RequiredArgsConstructor
+public class CustomApplicationRunner implements ApplicationRunner {
+    private final DataFetcher dataFetcher;
+
+    @Override
+    public void run(ApplicationArguments args) {
+        dataFetcher.populateCharactersDB();
+    }
+}
```
andrii-hoienko:
```suggestion
@Component
@RequiredArgsConstructor
public class DataPopulator implements ApplicationRunner {//Not best name, you can think about better one
    private final DataFetcher dataFetcher;
    private final Repo repo

    @Override
    public void run(ApplicationArguments args) {
        characters = dataFetcher.populateCharactersDB();
        repo.saveAll(characters)
    }
}
```

andrii-hoienko (changes requested):
Nice, left few comments

andrii-hoienko (approved):
Good job!

PR #186, opened 2024-08-28
jv-rick-and-morty solution

c-kalenyk:


src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
```diff
@@ -0,0 +1,35 @@
...
+    @Operation(summary = "Get random character", description = "Get information "
+            + "about random character")
+    @GetMapping("/random")
+    public Character getRandomCharacter() {
+        return characterService.getRandomCharacter();
+    }
+
+    @Operation(summary = "Search characters", description = "Get all characters "
+            + "whose names include requested part")
+    @GetMapping("/search")
```
okuzan, line 31:
```suggestion
    @GetMapping
```

okuzan: more RESTful, HTTP verb should be enough

okuzan (approved):
well done

PR #197, opened 2024-10-13
Do Rick and Morty Task

yurlis:


pom.xml
```diff
@@ -37,18 +39,127 @@
 			<artifactId>spring-boot-starter-data-jpa</artifactId>
 		</dependency>
 
+<!--		For Web controllers -->
```
Elena-Bruyako: remove comments and commented code

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,72 @@
...
+import mate.academy.rickandmorty.dto.RickAndMortyResponseDataDto;
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import mate.academy.rickandmorty.model.Character;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class CharacterServiceImpl implements CharacterService {
+    private static final String RICK_AND_MORTY_API_URL = "https://rickandmortyapi.com/api/character";
```
Elena-Bruyako: you can store this value in application proprties and get from it by using `@Value`

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,72 @@
...
+                        + randomNumber));
+    }
+
+    @Override
+    public List<Character> findByNameContaining(String name) {
+        return characterRepository.findByNameContainingIgnoreCase(name);
+    }
+
+    @Override
+    public void syncCharactersFromExternalApi() {
```
Elena-Bruyako: it's better to replace this logic to CharacterClient service for example

Elena-Bruyako (changes requested):

pom.xml
yurlis: Okay!  I specifically made comments for future projects because adding or incorrectly adding dependencies makes strange mistakes. It's easy to correctly add a dependency to a new project by looking at the training project. I'll save a copy to my local storage.

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
yurlis: No problem . I wanted to emphasise for those watching the project that there is an initialisation phase. That's why I did it.

liliia-ponomarenko (approved):
Awesome 😎

PR #198, opened 2024-10-18
Rick And Morty 1.0

aksoli666:
created api which contains two methods:
getRandomCharacter and getCharacterByNameContaining/ignore case/

### DIFF
--- a/src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
+++ b/src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
@@ -1,0 +1,38 @@
+package mate.academy.rickandmorty.controller;
+
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.tags.Tag;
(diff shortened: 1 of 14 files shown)


src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
```diff
@@ -0,0 +1,38 @@
...
+    public Character getRandomCharacter() {
+        return characterService.getRandomCharacter();
+    }
+
+    @Operation(
+            summary = "Get all characters by part of the name",
+            description = "Retrieve character catalog by part of the name and ignore case"
+    )
+    @GetMapping("/by-part-name")
+    public List<Character> getCharactersByNameContaining(String name) {
```
okuzan, line 35:
```suggestion
    @GetMapping
    public List<Character> getCharactersByNameContaining(@RequestParam String name) {
```
let's make it more RESTful

okuzan (approved):
Well done

PR #200, opened 2024-10-25
Solution

Kalixone:


src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
```diff
@@ -0,0 +1,35 @@
...
+    private final CharacterService characterService;
+
+    @GetMapping("/random")
+    @Operation(summary = "Get a random character",
+            description = "Returns a random character from the database.")
+    public Optional<Character> getRandomCharacter() {
+        return characterService.getRandomCharacter();
+    }
+
+    @GetMapping("/search")
```
George-Khomiak: @GetMapping
REST is rather resource oriented

src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
```diff
@@ -0,0 +1,35 @@
...
+@RequiredArgsConstructor
+@RequestMapping(value = "/api/characters")
+@Tag(name = "Character API", description = "Operations related to characters from Rick and Morty")
+public class CharacterController {
+    private final CharacterService characterService;
+
+    @GetMapping("/random")
+    @Operation(summary = "Get a random character",
+            description = "Returns a random character from the database.")
+    public Optional<Character> getRandomCharacter() {
```
George-Khomiak: Controllers should work with DTO objects, not with model. Create a response DTO for this endpoint

src/main/java/mate/academy/rickandmorty/controller/CharacterController.java
```diff
@@ -0,0 +1,35 @@
...
+        return characterService.getRandomCharacter();
+    }
+
+    @GetMapping("/search")
+    @Operation(summary = "Search characters by name",
+            description = "Returns a list of characters whose name contains the specified word.")
+    public List<Character> searchCharactersByName(@RequestParam String word) {
```
George-Khomiak: Same here, return list of DTO responses

src/main/java/mate/academy/rickandmorty/model/RickAndMortyResponse.java
```diff
@@ -0,0 +1,13 @@
+package mate.academy.rickandmorty.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class RickAndMortyResponse {
```
George-Khomiak: This class name is just the name of whole app, it's not clear why RickAndMorty fetches all characters, it's more like CharacterListDto or something like that

George-Khomiak (changes requested):
Add DTOs for your responses and a mapper if needed and we are good:)

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+            List<Character> characters = response.getResults()
+                    .stream()
+                    .map(result -> {
+                        Character character = new Character();
+                        character.setExternalId(result.getId());
+
+                        if (result.getName() != null) {
+                            character.setName(result.getName());
+                        } else {
+                            throw new IllegalArgumentException("Field name is missing");
```
slade13, line 56: Null-name guard is good, but throwing IllegalArgumentException will kill the entire startup—instead you might log and skip malformed entries to complete the import.

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+
+            characterRepository.saveAll(characters);
+
+            page++;
+        }
+    }
+
+    @Override
+    public CharacterResponseDto getRandomCharacter() {
+        int randomId = random.nextInt(1, (int) characterRepository.count());
```
slade13, line 72: `nextInt(1, N)` returns values in [1, N), so you’ll never pick the character whose ID equals the total count.

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+        Character character = characterRepository
+                .findById(Long.valueOf(randomId))
```
slade13, line 74: IDs in your DB may not be contiguous (deleted rows, failed imports), so an ID could be missing and trigger the “Character not found” exception.

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
slade13, line 72: You can do it for example like this:
```
List<Long> ids = characterRepository.findAllIds();       // custom repo method
long randomId = ids.get(random.nextInt(ids.size()));
return findById(randomId);
```

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+                .orElseThrow(() -> new RuntimeException("Character not found"));
+        return toDto(character);
+    }
+
+    @Override
+    public List<CharacterResponseDto> searchCharactersByName(String word) {
+        return characterRepository.findAll().stream()
```
slade13, line 81: In-memory filtering loads every Character into Java, which will choke as your dataset grows.
Move filtering into SQL:
```
List<Character> results = characterRepository.findByNameContainingIgnoreCase(word);
```

In your repo add:
```
List<Character> findByNameContainingIgnoreCase(String name);
```

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+
+@Service
+@RequiredArgsConstructor
+public class CharacterServiceImpl implements CharacterService {
+    private final RestTemplate restTemplate;
+    private final CharacterRepository characterRepository;
+    private final Random random;
+
+    @Override
+    public void fetchCharacters() {
```
slade13, line 25:
```suggestion
    @Transactional
    public void fetchCharacters() {
```

slade13 (approved):
Good job, please review my comments.

PR #215, opened 2025-01-04
First solution

mateuszwojtkowiak:


src/main/java/mate/academy/rickandmorty/service/RickAndMortyClient.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public CharacterDto getRandomCharacter() {
+        Transaction transaction = null;
+        int random = (int) (Math.random() * RANGE) + MIN;
+        HttpClient httpClient = HttpClient.newHttpClient();
+        String url = BASE_URL + CHARACTER + "/" + random;
+        HttpRequest httpRequest = HttpRequest.newBuilder().GET().uri(URI.create(url)).build();
+        try {
+            Session session = sessionFactory.openSession();
```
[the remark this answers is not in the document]
George-Khomiak: I think you can use finally to ensure session is closed properly, or try with resources

src/main/java/mate/academy/rickandmorty/service/RickAndMortyClient.java
```diff
@@ -0,0 +1,77 @@
...
+            transaction = session.beginTransaction();
+            HttpResponse<String> httpResponse = httpClient
+                    .send(httpRequest, HttpResponse.BodyHandlers.ofString());
+            ObjectMapper objectMapper = new ObjectMapper();
+            CharacterDto characterDto = objectMapper.readValue(httpResponse.body(),
+                    new TypeReference<>() {});
+            session.persist(characterDto);
```
George-Khomiak: Yeah, DTO is not something we want to save in the database, your Entity is Character so you should create a mapper and map your DTO to Character

src/main/java/mate/academy/rickandmorty/service/RickAndMortyClient.java
```diff
@@ -0,0 +1,77 @@
...
+    public CharacterDto getRandomCharacter() {
+        Transaction transaction = null;
+        int random = (int) (Math.random() * RANGE) + MIN;
+        HttpClient httpClient = HttpClient.newHttpClient();
+        String url = BASE_URL + CHARACTER + "/" + random;
+        HttpRequest httpRequest = HttpRequest.newBuilder().GET().uri(URI.create(url)).build();
+        try {
+            Session session = sessionFactory.openSession();
+            transaction = session.beginTransaction();
+            HttpResponse<String> httpResponse = httpClient
```
George-Khomiak: You are still working with API and this is not how this project should be done. The reqiurement is to fetch all data at the application start. So create a method like "fetchAllCharacters" and download all characters to your local database first. There is a few ways to achieve this. I mentioned an annotation in the chat that might be useful, but you can use other solutions you come up with or find in internet.

George-Khomiak (changes requested):
Left a few comments. Don't forget what DTOs are for, map them to Entity if you want to save it to database, create "fetchAll" or similar method and run it when the application starts. Rest of the methods should work with database, not API.

src/main/java/mate/academy/rickandmorty/controller/RickAndMortyController.java
```diff
@@ -18,13 +18,13 @@ public RickAndMortyController(RickAndMortyClient rickAndMortyClient) {
...
+    public CharacterModel getRandomCharacter() {
         return rickAndMortyClient.getRandomCharacter();
     }
 
     @Operation(summary = "Get list of characters matching parameter")
     @GetMapping("/character")
-    public List<CharacterDto> getCharactersList(@RequestParam String character) {
-        return rickAndMortyClient.getCharactersList(character);
+    public List<CharacterModel> getCharactersList(@RequestParam String name) {
+        return rickAndMortyClient.getCharactersList(name);
```
MateuszRostek: The parameter given here should be used to find all characters whose name contains a fragment, so the name can be like this: "nameSegment"

The name of this method could also be a bit clearer

src/main/java/mate/academy/rickandmorty/controller/RickAndMortyController.java
MateuszRostek, line 1: You should not change the type returned in the controller. The previous mentor mentioned that you should not write the DTO to the database, only the actual model. Controllers should return the DTO as they have been doing so far

src/main/java/mate/academy/rickandmorty/mapper/CharacterModelMapper.java
```diff
@@ -0,0 +1,15 @@
+package mate.academy.rickandmorty.mapper;
+
+import mate.academy.rickandmorty.dto.CharacterModelDto;
+import mate.academy.rickandmorty.model.CharacterModel;
+
+public class CharacterModelMapper {
```
MateuszRostek: You learned in this course how to use Mapstruct, use it here instead of writing your own mappers

src/main/java/mate/academy/rickandmorty/service/RickAndMortyClient.java
```diff
@@ -27,44 +31,68 @@ public RickAndMortyClient(SessionFactory sessionFactory) {
...
             }
             throw new RuntimeException(e);
         }
     }
 
-    public List<CharacterDto> getCharactersList(String character) {
+    @PostConstruct
+    public List<CharacterModelDto> fetchAllCharacters() {
+        StringBuilder builder = new StringBuilder();
+        for (int i = MIN; i <= MAX; i++) {
```
MateuszRostek: You should not statically save the MIN MAX values, because if there are any changes in the external API, your code will become deprecated.

MateuszRostek: For example, you can create a loop that executes until the “next” field is null

Zrzut ekranu 2025-01-09 195712

MateuszRostek (changes requested):
Please see some comments and implement them:)

src/main/java/mate/academy/rickandmorty/mapper/CharacterModelMapper.java
```diff
@@ -0,0 +1,10 @@
+package mate.academy.rickandmorty.mapper;
+
+import mate.academy.rickandmorty.dto.CharacterModelDto;
+import mate.academy.rickandmorty.model.CharacterModel;
+
+public interface CharacterModelMapper {
```
slade13, line 6: The interface `CharacterModelMapper` should be annotated with `@Mapper` from MapStruct if you intend to use MapStruct for mapping. This will enable automatic implementation generation.

src/main/java/mate/academy/rickandmorty/controller/RickAndMortyController.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Operation(summary = "Get random character")
+    @GetMapping
+    public CharacterModelDto getRandomCharacter() {
+        return rickAndMortyClient.getRandomCharacter();
+    }
+
+    @Operation(summary = "Get list of characters matching parameter")
+    @GetMapping("/character")
+    public List<CharacterModelDto> getCharactersList(@RequestParam String nameSegment) {
```
slade13, line 27: The RequestParam annotation should have a more descriptive name. Consider renaming 'nameSegment' to 'characterName' for better clarity.

src/main/java/mate/academy/rickandmorty/controller/RickAndMortyController.java
```diff
@@ -0,0 +1,30 @@
...
+@RestController
+public class RickAndMortyController {
+    private final RickAndMortyClient rickAndMortyClient;
+
+    public RickAndMortyController(RickAndMortyClient rickAndMortyClient) {
+        this.rickAndMortyClient = rickAndMortyClient;
+    }
+
+    @Operation(summary = "Get random character")
+    @GetMapping
```
slade13, line 20: The GetMapping annotation for getRandomCharacter method should specify a path. Consider using a specific endpoint like '/character/random' to avoid potential conflicts with other endpoints.

src/main/java/mate/academy/rickandmorty/mapper/CharacterModelMapperImpl.java
```diff
@@ -0,0 +1,33 @@
...
+                characterModel.getStatus());
+    }
+
+    @Override
+    public CharacterModel toModel(CharacterModelDto characterModelDto) {
+        if (characterModelDto == null) {
+            return null;
+        }
+        CharacterModel characterModel = new CharacterModel();
+        characterModel.setExternalId(characterModelDto.id());
```
slade13, line 27: The method `id()` is used to access the `externalId` field in `CharacterModelDto`, which suggests that `CharacterModelDto` might be using records or a similar structure. Ensure that the method name matches the field name in `CharacterModelDto`. If `externalId` is the field name, use `characterModelDto.externalId()` instead.

slade13 (approved):
Good job on implementing requested changes, please review my comments.

PR #236, opened 2025-03-28
Fetch data feature

miktarn:


src/main/resources/application.properties
```diff
@@ -1 +1,6 @@
-
+spring.datasource.url=jdbc:mysql://localhost:3307/characters?useSSL=false&serverTimezone=UTC
+spring.datasource.driverClassName=com.mysql.cj.jdbc.Driver
+spring.datasource.username=admin
+spring.datasource.password=root
+spring.jpa.database-platform=org.hibernate.dialect.MySQL8Dialect
+spring.jpa.hibernate.ddl-auto=create-drop
```
d1sam, line 6: remove real creds from repo, instead use env-variables.

src/main/java/mate/academy/rickandmorty/dto/CharacterMapper.java
d1sam, line 1: Why don't use MapStruct?

src/main/java/mate/academy/rickandmorty/repository/CharacterRepository.java
```diff
@@ -0,0 +1,12 @@
+package mate.academy.rickandmorty.repository;
+
+import java.util.List;
+import mate.academy.rickandmorty.model.Character;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
```
d1sam, line 8: not necessary, but it's minor

src/main/java/mate/academy/rickandmorty/dto/CharacterMapper.java
miktarn: Because there only two methods for mapping so there is no big need for MapStruct

src/main/java/mate/academy/rickandmorty/service/impl/CharacterServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+public class CharacterServiceImpl implements CharacterService {
+    private final CharacterRepository characterRepository;
+
+    @Override
+    public Character getRandomCharacter() {
+        long charactersAmount = characterRepository.count();
+        if (charactersAmount == 0) {
+            throw new RuntimeException("Not enough characters to process");
+        }
+        long randomId = new Random().nextLong(1, charactersAmount);
```
d1sam, line 23: you will create new instance of Random on each method call.

d1sam (changes requested):
minor changes should be done.

PR #278, opened 2026-01-18
Random character wiki

koder95:
Introduces model, DTOs (internal and external), enums for Gender and Status, a MapStruct mapper with conversion logic, and a mapper config. These additions enable mapping between external API data and internal representations for character information.

Introduced CharacterController, CharacterRepository, CharacterService, and ImportService to support character import, random retrieval, and search by name. Updated Application to trigger character import on startup.

Implemented CharacterServiceImpl with methods for saving, retrieving random characters, and searching by name. Added ImportServiceImpl to import characters from an external API, including error handling with a new InvalidConnectionException.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,6 +126,7 @@
 					<consoleOutput>true</consoleOutput>
 					<failsOnError>true</failsOnError>
 					<linkXRef>false</linkXRef>
+					<sourceDirectories>src</sourceDirectories>
 				</configuration>
 			</plugin>
 		</plugins>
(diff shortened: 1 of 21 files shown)


koder95:
@mateacademy-ai-mentor I reply to your [comment](https://github.com/mate-academy/jv-rick-and-morty/pull/278#discussion_r2702545507):
No. This method throws `tools.jackson.core.JacksonException` which extends `java.lang.RuntimeException`. So I think you are wrong.

PR #292, opened 2026-03-27
add rick-and-morty solution

RodionTarasov:


src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+            response = httpClient.send(
+                    httpRequest,
+                    HttpResponse.BodyHandlers.ofString()
+            );
+        } catch (IOException | InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+        if (response.statusCode() != 200) {
+            throw new RuntimeException("API вернул статус " + response.statusCode()
+                    + " для URL: " + url);
```
fetis-off, line 85: Use only english language in the code

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+import mate.academy.rickandmorty.dto.ExternalCharacterDto;
+import mate.academy.rickandmorty.dto.RickAndMortyPageDto;
+import mate.academy.rickandmorty.mapper.CharacterMapper;
+import mate.academy.rickandmorty.model.CharacterEntity;
+import mate.academy.rickandmorty.repository.CharacterRepository;
+import org.springframework.stereotype.Service;
+
+@RequiredArgsConstructor
+@Service
+public class CharacterServiceImpl implements CharacterService {
```
fetis-off, line 23: I see an Single Responsibility violation in your `CharacterService`.
I would rather create a separate class and called id `RickAndMortyClient` which is responsible for request to the API. Your `CharacterService` should only work with `Character` entities and has basic methods like (save, saveAll, etc)

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+            try {
+                Thread.sleep(250);
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+            currentUrl = pageDto.getInfo().getNext();
+        }
+    }
+
+    private void saveToDatabase(List<ExternalCharacterDto> pageDtoResults) {
```
fetis-off, line 62:
```suggestion
    @Transactional
    public void saveToDatabase(List<ExternalCharacterDto> pageDtoResults) {
```

src/main/java/mate/academy/rickandmorty/service/CharacterServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+        for (ExternalCharacterDto ecd : pageDtoResults) {
+            characterRepository.save(characterMapper.toModel(ecd));
+        }
+    }
+
+    private RickAndMortyPageDto fetchPage(HttpClient httpClient, String url) {
```
fetis-off, line 68:
```suggestion
```
This method should not be in `CharacterService`

src/main/java/mate/academy/rickandmorty/mapper/CharacterMapperImpl.java
```diff
@@ -0,0 +1,21 @@
+package mate.academy.rickandmorty.mapper;
+
+import mate.academy.rickandmorty.dto.ExternalCharacterDto;
+import mate.academy.rickandmorty.model.CharacterEntity;
+import org.springframework.stereotype.Component;
+
+@Component
+public class CharacterMapperImpl implements CharacterMapper {
```
fetis-off, line 8: Why did not you use Mapstruct for mapper?

fetis-off (changes requested):
Image

Did you test it locally? Because there is an exception so build is failed and not all characters were added to database. You should make a second break 1 second between requests (now you have 250 ms)

